PR #13, opened 2026-07-17
feat: add COLLIE_SKIP_SERVE to bypass tailscale serve

diogenesc:
## Summary

Adds `COLLIE_SKIP_SERVE=1` env var to skip `tailscale serve` entirely. The bridge stays on loopback only (or `0.0.0.0` if `COLLIE_HOST` is set), ideal for deployments behind a reverse proxy (Caddy, Nginx, etc.) or direct port access via MagicDNS.

## Changes

- `scripts/collie-ctl.sh`: `cmd_serve()`, `cmd_unserve()`, `cmd_status()`, and `print_status_banner()` all respect `COLLIE_SKIP_SERVE=1`
- `.env.example`: documents the new variable
- Version bumped to 0.12.0 (minor)

## Use case

When running Collie behind Caddy/Nginx or accessing directly via `http://host:port`, there's no need for `tailscale serve` — your proxy handles TLS, auth, and public access. This avoids the bridge fighting over the tailscale serve mapping.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,6 +12,10 @@ COLLIE_HOST=127.0.0.1
 # (plain HTTP on :$COLLIE_PORT — for Headscale / `.internal` domains without HTTPS certs; then set
 # COLLIE_PUBLIC_HOSTS below, and note PWA install + Web Push need a secure context).
 # COLLIE_SERVE_MODE=https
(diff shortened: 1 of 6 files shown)


AltanS:
Thanks @diogenesc! this is a use case I want to support properly, so I picked it up into #14 and bundled the security-docs changes it needed; your commit rides along there. It'll close this PR when it lands. Appreciate it 🙏

PR #17, opened 2026-07-20
fix: preserve Herdr snapshot tab order

iFwu:
## Summary

Preserve the tab order reported by Herdr from the bridge snapshot through the space grouping and tab-strip render paths.

Herdr can reorder tabs without renumbering their stable public numbers. Collie currently sorts `TabView` values by `number` in three layers, so a user-visible reorder is changed back to creation/number order before it reaches the screen.

## Reproduction

1. Open a Herdr workspace with at least two tabs.
2. Reorder the tabs in Herdr so the snapshot order differs from their stable numbers.
3. Open the same workspace or pane in Collie.

A minimal equivalent snapshot is:

```text
[t2 (number 2), t1 (number 1)]
```

Before this change, Collie projects and renders it as `[t1, t2]`. After this change, Collie keeps `[t2, t1]`, matching Herdr.

## Change

- Stop sorting tabs by `number` in the state engine.
- Keep the incoming order when grouping a workspace's panes by tab.
- Keep the incoming order in `TabStrip`.

Workspace ordering, agent priority ordering, pane ordering, tab IDs, and the `number` field itself are unchanged.

## Verification

The regression assertions failed before the production change in all three affected layers:

- state-engine snapshot projection
- workspace tab grouping
- rendered tab strip

After the change:

- `bun run typecheck` (root): pass
- `bun run test` (bridge): 219 pass
- `bun run typecheck` (web): pass
- `bun run test` (web): 906 pass
- `bash scripts/check-version.sh`: pass (`0.13.1`)

### DIFF
--- a/bridge/state-engine.test.ts
+++ b/bridge/state-engine.test.ts
@@ -189,6 +189,18 @@ describe("StateEngine — in-flight guard", () => {
 });
 
 describe("StateEngine — snapshot shaping", () => {
+  test("preserves the tab order reported by Herdr", async () => {
+    const { herdr, engine, poll } = makeEngine();
+    herdr.tabs = [
+      { ...herdr.tabs[0]!, tab_id: "w1:t2", number: 2, label: "second" },
+      { ...herdr.tabs[0]!, tab_id: "w1:t1", number: 1, label: "first" },
+    ];
+
+    await poll();
+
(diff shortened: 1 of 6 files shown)


AltanS:
Thanks! 👍

PR #18, opened 2026-07-20
fix: focus composer within mobile tap

iFwu:
## Summary

Focus the composer synchronously when the user taps raw terminal output, keeping the call inside the tap's user-activation window so mobile browsers can open the software keyboard.

## Reproduction

1. Open a pane in the Collie PWA on iOS.
2. Tap raw text in the terminal mirror instead of tapping the composer directly.
3. Observe that the composer may become focused after the tap, but the software keyboard does not open.
4. Tap the composer itself and observe that the keyboard opens normally.

The mirror path currently defers `focus()` with `setTimeout(0)`. Mobile browsers may no longer treat that callback as part of the user gesture, so programmatic focus cannot summon the keyboard.

## Change

`ComposerHandle.focusInput`, which is called by the terminal-mirror click handler, now focuses and positions the caret synchronously.

The existing deferred helper remains in place for internal flows that update the composer value first, including Take over, command insertion, and image upload. Interactive controls inside the mirror and text-selection taps keep their existing focus guards.

## Verification

A regression test now asserts that the textarea is focused before `fireEvent.click()` returns. It failed before the production change and passes afterward.

- `bun run typecheck` (root): pass
- `bun run test` (bridge): 218 pass
- `bun run typecheck` (web): pass
- `bun run test` (web): 906 pass
- `bash scripts/check-version.sh`: pass (`0.13.1`)

### DIFF
--- a/web/src/components/agent-chat.test.tsx
+++ b/web/src/components/agent-chat.test.tsx
@@ -1,5 +1,5 @@
 import { useState, type ComponentProps } from "react";
-import { act, render, screen, waitFor } from "@testing-library/react";
+import { act, fireEvent, render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { http, HttpResponse } from "msw";
 import { createMemoryRouter, RouterProvider, useParams } from "react-router";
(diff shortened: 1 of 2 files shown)


AltanS:
Thanks! 👍

PR #26, opened 2026-07-26
feat: add managed NetBird front door

iamtimmy:
## Summary
- add `COLLIE_FRONT_DOOR=tailscale|netbird|proxy` while retaining `COLLIE_SKIP_SERVE=1` compatibility
- add authenticated `netbird expose` management with systemd supervision and a PID-backed fallback
- make Tailscale and NetBird cutovers fail closed around handler ownership, process identity, and teardown errors
- add permanent shell lifecycle coverage and focused deployment/security documentation

## Testing
- [x] `bash scripts/check-version.sh`
- [x] `bun run typecheck`
- [x] `bun run test`
- [x] `cd web && bun run typecheck`
- [x] `cd web && bun run test`
- [x] `bash -n scripts/collie-ctl.sh scripts/collie-ctl.test.sh`

## Notes
- keeps the existing `0.14.2` version/changelog state; this PR does not cut or tag a release
- preserves the original Tailscale-only package and plugin descriptions

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,24 +1,49 @@
 # Collie configuration. Copy to your plugin config dir as `.env`:
 #   cp .env.example "$(herdr plugin config-dir herdr.collie)/.env"
-# All values are optional; the defaults suit a single-user, tailnet-only setup.
+# Values are optional; the default selects the Tailscale front door. Choose exactly one front door
+# before the first `start`: Tailscale, NetBird, or your own authenticated proxy.
 
 # --- Networking ---
-# Port the bridge listens on (loopback only; tailscale serve proxies to it).
+# Port the bridge listens on (loopback only; the selected front door proxies to it).
 COLLIE_PORT=8787
 # Bind host. Keep this loopback. Binding 0.0.0.0 makes identity checks meaningless.
 COLLIE_HOST=127.0.0.1
+# Which front door collie-ctl.sh manages: "tailscale" (default), "netbird", or "proxy".
+# "proxy" starts only the loopback bridge; your Caddy/Nginx/etc. owns TLS, auth, and public access.
+# Starting or switching tracks retires stale managed Tailscale/NetBird ingress; proxy retires both.
+# COLLIE_FRONT_DOOR=tailscale
+#
 # How the bridge is published on the tailnet (read by collie-ctl.sh when it runs `tailscale serve`,
 # not by the bridge itself): "https" (default — tailnet :443, Tailscale-managed cert) or "http"
 # (plain HTTP on :$COLLIE_PORT — for Headscale / `.internal` domains without HTTPS certs; then set
 # COLLIE_PUBLIC_HOSTS below, and note PWA install + Web Push need a secure context).
 # COLLIE_SERVE_MODE=https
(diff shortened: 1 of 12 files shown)


AltanS:
Thanks for this, it's careful work, and the Tailscale ownership tracking in it is a real improvement I want to keep.

I'm declining the NetBird front door, and that's not about the quality of the patch. Collie manages exactly one front door because it's the one I actually run: `tailscale serve` is on the deployment host, so it's testable and breakage shows up the same day. NetBird I'd be shipping blind. It isn't installed here, there's no CI for it, and shell code sits outside both test suites, so its CLI contract (flag names, the v0.66 floor, what `expose` publishes) becomes something I maintain by reading a description. It's also a precedent I can't hold up evenly, since Cloudflare Tunnel, ZeroTier and Twingate all have equal claim.

You're not blocked, though. `COLLIE_SKIP_SERVE=1` plus `netbird expose 8787` is the whole integration. The PR adds supervision and teardown on top of that, which is the same thing I deliberately don't do for anyone's Caddy under Variant C.

What the PR does point at fairly is that the README makes Tailscale look mandatory when it's only the default. I'm fixing that with a generic "any other mesh or tunnel" variant covering the proxy requirements, why `COLLIE_TRUSTED_USER` is inert there, and `COLLIE_DEVICE_HEADER` when your tunnel injects one. Review welcome.

Two bugs worth having either way.

**The ownership tracking bricks existing installs.** `ensure_tailscale_root_available` refuses to publish over an existing root mount, and `stop_tailscale_serve` won't remove one that predates the ownership file, which is the state of every deployment alive today. Against my host's real serve config:

```
tailscale serve: no Collie-managed mapping recorded
error: Tailscale serve already has an unowned root mount on :8787; refusing to overwrite it
```

`cmd_serve` returns 1 and `cmd_start` dies on `set -e`, so `start`, `restart` and `update` all fail. Fix is adopt-on-match: if the existing root proxies to exactly `http://127.0.0.1:$PORT` on the matching protocol, claim it and continue. Your suite covers the foreign-mount case correctly, it just never covers "this mount is Collie's own, from before ownership existed".

**NetBird credentials land in world-readable argv.** The runner passes `--with-pin` / `--with-password` on the command line, so `ps -eo args` and `/proc/<pid>/cmdline` (mode 444) hand them to any local user:

```
bash …/netbird expose 8787 --with-name-prefix collie --with-pin 482913 --with-password hunter2…
```

Those credentials are the only thing between a public URL and a shell, so worth fixing wherever this lands.

I'm keeping the ownership tracking. It fixes a real bug in main, where `unserve` blindly runs `tailscale serve --https=443 off` and can remove a mapping Collie never created. It'll land as its own commit authored by you, plus the adopt-on-match fix as a separate commit on top, minus the `COLLIE_FRONT_DOOR` switch since that collapses to a rename of `COLLIE_SKIP_SERVE` once NetBird is out. Your Variant D number collided with something I shipped in 0.15.0 after you opened this, not your doing.

AltanS:
Superseded by #36, which lands the Tailscale ownership tracking with you as commit author, plus the adopt-on-match fix so existing installs survive the upgrade, and Variant E for the bring-your-own-tunnel case. Shipped in 0.16.0. Thanks again for the work, and for the shell test harness in particular.

PR #27, opened 2026-07-26
feat: Windows support for the bridge via herdr's named pipe

mikebenner:
## What

Runs the Collie bridge natively on Windows against herdr's Windows beta. One new file (`bridge/dial.ts`) plus routing the two `Bun.connect` sites in `bridge/herdr-client.ts` through it. No behavior change on Linux/macOS — off win32 the shim delegates straight to `Bun.connect({unix})`.

## Why it works

herdr's Windows beta doesn't use AF_UNIX: the `.sock` path on disk is a small pointer file (`<pid>:<start_ns>`), and the actual transport is a **named pipe whose name is the full socket path** — `\\.\pipe\C:\Users\<you>\AppData\Roaming\herdr\herdr.sock`. `Bun.connect({unix})` can't dial named pipes, but Bun's `node:net` can, and it adapts cleanly to the same `write`/`flush`/`end` handler shape both call sites already use (the settled/fireDown re-entrancy guards work unchanged; the shim maps `end()` to `destroy()` so close stays prompt for the one-shot RPCs).

## Scope / limitations

- Bridge only. `collie-ctl.sh`, the systemd unit, and `tailscale serve` are untouched — on Windows I run `bun run bridge/index.ts` directly, in the README's Variant C posture (loopback bind, own ingress in front, `COLLIE_PUBLIC_HOSTS` pinned).
- Multi-session discovery isn't adapted (POSIX path derivation), so `COLLIE_MULTI_SESSION=off` is the right setting on Windows.
- `herdr-plugin.toml` platforms are untouched (the plugin actions are bash).

## Tested

- Live against herdr `0.7.4-preview` (protocol 16) on Windows 11: `session.snapshot` polling, `events.subscribe` stream up + resubscribe on pane changes, pane reads and replies, web UI served — running as my daily driver behind a NetBird mesh.
- `bun run build` (both typechecks) passes on Windows.
- `bun run test`: 205 pass / 14 fail — **identical counts on a pristine checkout of `afa53fe`**; the 14 are pre-existing Windows-environment assumptions in tests (0600 permission fixtures, POSIX path fixtures), none in code this PR touches.

## Versioning

Left the version/CHANGELOG for your release flow (matching how #22 landed) — happy to push a `0.15.0` bump + CHANGELOG entry citing the feature hash if you'd rather have it in the PR, and likewise a short README "Windows (experimental)" note.

https://claude.ai/code/session_01VutEkP9A8oY72ZQBfFJdSw

### DIFF
--- a/bridge/config.test.ts
+++ b/bridge/config.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { join } from "node:path";
 
-import { loadConfig } from "./config.ts";
+import { defaultSocketPath, loadConfig } from "./config.ts";
 
 // loadConfig is the deployment contract — env vars in, a resolved Config out. Pure (just reads
 // process.env + homedir), so we drive it by mutating the environment and restoring it after.
@@ -176,5 +177,26 @@ describe("loadConfig", () => {
     const cfg = loadConfig();
     expect(cfg.trustedUser).toBe("me@example.com");
     expect(cfg.host).toBe("0.0.0.0");
+  });
+});
+
+// Pure — both platform branches are testable from any host (expectations use join() so the
+// host's separator never leaks into the assertion).
+describe("defaultSocketPath", () => {
+  test("unix default lives under ~/.config/herdr", () => {
+    expect(defaultSocketPath("linux", {}, "/home/u")).toBe(join("/home/u", ".config", "herdr", "herdr.sock"));
+    expect(defaultSocketPath("darwin", {}, "/Users/u")).toBe(join("/Users/u", ".config", "herdr", "herdr.sock"));
+  });
+
+  test("win32 default honours APPDATA", () => {
+    expect(defaultSocketPath("win32", { APPDATA: "C:\\Users\\u\\AppData\\Roaming" }, "C:\\Users\\u")).toBe(
+      join("C:\\Users\\u\\AppData\\Roaming", "herdr", "herdr.sock"),
+    );
+  });
+
+  test("win32 falls back to <home>/AppData/Roaming when APPDATA is unset", () => {
+    expect(defaultSocketPath("win32", {}, "C:\\Users\\u")).toBe(
+      join("C:\\Users\\u", "AppData", "Roaming", "herdr", "herdr.sock"),
+    );
   });
 });
--- a/bridge/config.ts
+++ b/bridge/config.ts
@@ -139,6 +139,23 @@ export interface Config {
   skipServe: boolean;
 }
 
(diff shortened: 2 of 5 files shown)


mikebenner:
Pushed 85a7411 addressing the gaps flagged in the #25/#27 cross-review: `onDial` cancellation so a timeout mid-connect can't leak a pending dial (wired into both `request()` and `subscribeEvents()`), `toPipeName()` normalization for already-prefixed pipe names, a win32 `%APPDATA%\herdr` default so `HERDR_SOCKET_PATH` is no longer required, and a new `dial.test.ts` with live named-pipe coverage on Windows (round-trip, mid-codepoint chunk split, connect error, cancel) — pure parts run everywhere, pipe tests skip off win32. Suite: 214 pass / same 14 pre-existing Windows fixture failures as `main`. Still deliberately bridge-only; happy to rebase onto or combine with #25's packaging if that's the direction you choose.

https://claude.ai/code/session_01VutEkP9A8oY72ZQBfFJdSw

AltanS:
Thanks, this is the right shape. One shim, one seam, `HerdrClient` untouched. The pointer-file plus full-path pipe name detail is what made the competing diagnosis in #25 make sense, and it matches the `interprocess` name mapping in the crate source.

Your `onDial(cancel)` also fixes a real pre-existing bug on Linux and macOS, not just Windows. A request that timed out mid-connect had no handle to `end()` and leaked the pending dial. That would have been worth a PR on its own.

Both your commits are cherry-picked onto main with authorship preserved.

What I added on top:

- `COLLIE_HERDR_DIAL=auto|net|bun`, from bwright's PR. Forcing `net` runs your win32 branch on POSIX, since `net.connect(path)` takes an AF_UNIX path there. Your four live-dial tests were `describe.skip` off win32, so the code Windows depends on was unexercised on any machine that could review it. They now run everywhere, 6 pass and 0 skip, and sabotaging `net.connect` fails them.
- Verified end to end against a live herd, both dialers, protocol 17: identical snapshots and `events.subscribe` acked over `node:net`.
- The packaging you scoped out, `.env.example`, a README "Windows (experimental)" section, and an ARCHITECTURE note.
- A note at the top of `dial.ts` recording that there is no handshake, sourced from the crate docs, so nobody re-derives it.

I left `herdr-plugin.toml` platforms alone, agreeing with your reasoning. The bridge runs on Windows, the launcher does not.

On your review of #25: it was accurate and well evidenced, and re-verifying the central claim before commenting was the right call. Thanks for doing that rather than just advocating for your own PR.

Merged to main as dd6610d. Closing this as merged via cherry-pick.

PR #41, opened 2026-07-28
Light and system themes

konpyl:
Collie has been hardcoded to dark since 0.1.0 — `class="dark"` on `<html>`, with a full light token
set sitting unused in `index.css` the whole time. This unpins it, adds a System mode that follows the
OS, and puts a control in two places.

Four commits, each readable on its own:

| | |
|---|---|
| `7a7a13e` | **A pre-existing bug**, split out — the space and tab chip rows painted over each other |
| `513d368` | The feature |
| `bb1f087` | ADR 0002 + the CLAUDE.md rule that links to it |
| `26736f1` | Release 0.18.0 |

### The interesting decision

The design originally gave the mirror a light ANSI palette — define `--ansi-0…15` twice, emit
`var(--ansi-N)` from `lib/ansi.ts`, let CSS swap them. Built, tested, and then measured against real
panes:

| pane | truecolor | 256-colour | basic | bright |
|---|---|---|---|---|
| w2Z:p1 | 446 | 7 | 0 | 0 |
| wJ:p1 | 150 | 6 | 0 | 0 |
| w14:p1 | 461 | 6 | 0 | 0 |
| w18:pD | 109 | 6 | 0 | 0 |

**Zero basic and zero bright ANSI codes.** Claude Code emits truecolor almost exclusively, and
truecolor names an absolute sRGB value — no palette can redirect it. On a white ground, eight of
thirteen distinct colours in a real pane fell below 3:1, including a `●` at **1.0:1** and Monokai's
foreground at 1.07:1.

So the mirror now renders in dark space under every theme and light inverts it with
`invert(1) hue-rotate(180deg)`, which preserves hue well enough to keep syntax highlighting legible.
Sampling rendered pixels, light tracks dark almost exactly — median 6.73 vs 7.46. The reasoning, the
three rules it imposes, and what would justify revisiting it are in
[ADR 0002](.adr/0002-invert-the-light-terminal-mirror.md).

`lib/ansi.ts` still emits `var(--ansi-N)` for indexed colour, and does so via **both** spellings —
`color256()` resolves 0–15 through the same table, so theming `31m` but not `38;5;1` would render the
same logical colour two different ways in one pane.

### Notes for review

- **No CSP change.** The anti-flash script is a same-origin file, which `script-src 'self'` already
  permits. System users get a correct first paint from CSS alone, with no JS at all.
- **One token block, not two.** Verified first that `color-mix()` resolves a `light-dark()` argument,
  since the app-wide `border-border/60` and `outline-ring/50` base rules depend on it.
- **Contrast figures are measured, not modelled** — rasterized through a canvas to resolve
  `oklch()`/`light-dark()`, composited against the real ancestor stack. Three rounds of
  browser-driven UX review across 390/768/1280 × both themes found 13 further defects (an unreadable
  off-switch, half-strength focus rings, every alpha-modified muted text under 3:1); two of those
  were regressions introduced by the previous round's fixes.
- **Known cost, deliberately not fixed:** diffs and inverse-video regions become dark slabs in light.
  Legibility survives; only visual weight inverts. Documented in the ADR — greying it is worse.
- **iOS status bar is unresolved.** `apple-mobile-web-app-status-bar-style: black-translucent` renders
  white glyphs, which will vanish on a light page in an installed PWA. There's no `media` form and no
  runtime API, and switching to `default` changes the layout contract with `viewport-fit=cover`. Needs
  a real device.
- **`26736f1` is droppable.** If you'd rather cut releases yourself, drop the release commit and the
  other three still apply cleanly.

`bun run build` clean; 1064 web tests and 318 root tests pass; `scripts/check-version.sh` ✓ at every
commit on the branch, not just the tip.

### DIFF
--- a/.adr/0002-invert-the-light-terminal-mirror.md
+++ b/.adr/0002-invert-the-light-terminal-mirror.md
@@ -1,0 +1,105 @@
+# 0002 — The light terminal mirror is inverted, not re-themed
+
+- **Status:** Accepted
+- **Date:** 2026-07-28
+- **Shipped in:** 0.18.0
+- **Trail:** measurements below were taken against live panes on 2026-07-28; the counts and contrast
+  figures are reproducible with the method each one names
+
+## Context
+
+Adding light mode meant deciding what the pane mirror does. The obvious answer, and the one this
+work carried for most of its life, was: give the mirror a light ANSI palette. Define
+`--ansi-0…15` twice — VS Code's Dark+ set for dark, its light-theme set for light — have
+`lib/ansi.ts` emit `var(--ansi-N)` instead of literal hex, and let CSS swap them. That was built and
+tested before anyone measured whether it would matter.
+
+It doesn't, much. Counting SGR forms in four live panes:
+
+| pane | truecolor (`38;2`) | 256-colour (`38;5`) | basic (`30–37`) | bright (`90–97`) |
+| --- | --- | --- | --- | --- |
+| w2Z:p1 | 446 | 7 | 0 | 0 |
+| wJ:p1 | 150 | 6 | 0 | 0 |
+| w14:p1 | 461 | 6 | 0 | 0 |
+| w18:pD | 109 | 6 | 0 | 0 |
+
+**Zero basic and zero bright codes anywhere.** Claude Code emits truecolor almost exclusively, and
+truecolor names an absolute sRGB value — there is no palette slot to redirect. The 16 themeable
+slots the design was built around are very nearly unreachable in practice.
+
+The consequence on a white ground was measured on a real pane: of 13 distinct rendered colours,
+eight fell below 3:1, including a `●` at **1.0:1** (`#ffffff` on white) and Monokai's foreground
+`#f8f8f2` at **1.07:1**. A light mirror built the obvious way is not a dimmer light mirror; it is an
+unreadable one.
+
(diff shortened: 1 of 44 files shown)


AltanS:
Superseded by #46, which is now merged.

Your four commits are carried **verbatim** and are in `main`'s history with your authorship intact — `636b7af`, `59bcfe1`, `78425bd`, `0c589bc`. Thanks for this; the measurement work behind ADR 0002 is what made the decision defensible.

Three things changed on top:

- **ADR 0002's rule 2 turned out to be false.** `color-scheme: dark` on an element *does* flip an inherited `light-dark()` token — resolution is element-scoped, per spec, verified in Chromium. Nothing rendered wrong, since the literals are byte-exact matches for those dark halves, but the rule is now stated as the convention it is rather than a platform constraint.
- **The colour-form table couldn't see codex.** It counted all `38;5;n` as 256-colour, hiding palette slots 0–15 inside the unthemeable cube — and codex emits `38;5;1`/`38;5;3`/`38;5;6` with zero basic codes. Re-measured per harness: opencode 100% truecolor, pi 89%, claude 79%, codex 7%. The conclusion holds for three of four, which is a stronger claim than the original generalisation from four Claude panes.
- **Appearance moved to Settings only**, and `0.19.0 → 0.20.0` since main took 0.19.0 for the journal adapters while this was open.

Closing in favour of #46.

PR #42, opened 2026-07-28
feat: triage the dashboard by attention and recency, and stop calling every agent "claude"

konpyl:
> **Stacked on #41.** This branch is cut from `feat/light-and-system-theme`, so until #41 merges the
> diff below also contains its four commits (0.18.0). Once #41 lands on `main` this PR's diff reduces
> to its own 15 commits on its own — no rebase needed, since the repo merges rather than squashes.
> **Please merge #41 first.** Reviewable now; the commits to read start at `feat(bridge): keep two
> timestamps per pane`.

## The problem

Open Collie on a phone with a real herd — 20 agents, 37 bare shells, 45 spaces — and the dashboard
answers a question nobody asked. It lists 45 spaces in workspace-creation order above everything
else; it sorts agents by `STATUS_RANK → workspaceNumber → paneId`, so nothing distinguishes an agent
that finished ten seconds ago from one you dealt with yesterday; and every row is titled `claude`,
because the title falls back to the agent name and the tab label — the one thing that says what the
work *is* — was never on screen at all.

## The idea

Herdr reports **no timestamps of any kind** — not on panes, tabs, or workspaces. So Collie derives
and owns them, and the whole feature falls out of **two numbers per pane**:

```
activeAt — when the agent last changed status   (the bridge observes it)
seenAt   — when you last opened or drove it     (through Collie, and only through Collie)
```

"Unseen" is then a **comparison, not a stored flag**: `status === "done" && activeAt > seenAt`. There
is no read-receipt table and nothing to keep in sync — opening the pane bumps `seenAt` past
`activeAt` and the row leaves the section by itself.

## What changes

**Ordering.** `lib/triage.ts` (replacing `lib/agent-groups.ts`) buckets the herd into
**Needs you → Ready · unseen → Working → Recent**. The first three are pinned: they never move,
never invert, and never fold — collapsing an alert defeats the alert. Recent runs by when you last
used each pane and is the only section the direction toggle reaches. Recent and Spaces fold and
remember it; fold both and the page is the triaged herd and nothing else.

**Spaces moved below every agent section.** It's a navigator, not a work queue, so it sits under the
thing it navigates to — recency-ordered, with a filter box, folded by default above 8 spaces.

**Naming.** A row is titled `project · tab` (`moonward_os · fix-auth`), rendered as two spans so the
*tab* survives truncation — eight panes in one project all begin `moonward_os · `, and truncating the
joined string left several rows rendering the identical `moonward_os · t…`. The pane's own name (a
herdr label, or Claude's `/rename` session name) moves to line 2. The agent's identity was always the
avatar, never the text.

**Tab and space chips carry a status dot** — blocked / ready / working / idle, in the herd list's own
palette. They previously showed a dot only for blocked, so every other state read identically.

**Reach.** The dashboard, the in-pane switcher and the space view share one classifier (`bucketOf`),
so a chip and a row can't disagree about what a colour means.

## Backward compatibility

An older bridge reports no timestamps, and the client degrades to today's dashboard **with no branch
and no feature detection**: the unseen test is `false`, every comparator returns `0`, and
`Array.prototype.sort` is stable — so each section keeps the bridge's existing order and
Ready · unseen is simply empty.

## Security

Marking a pane seen made a **read-level GET mutate server state**, which no read had done before —
and `checkAccess` deliberately doesn't require an `Origin` on reads (browsers omit it on same-origin
GETs). A page visited while on the tailnet could therefore have fired ``
at guessable pane ids and silently cleared the whole Ready · unseen section. Caught in this branch's
own security review and fixed before it shipped: `marksPaneSeen()` gates on a custom request header,
which a no-cors cross-site request cannot set without a preflight the bridge never answers. Write
actions need no header — they already clear the `Origin`-requiring write gate.

## Decision record

**[ADR 0003](.adr/0003-one-shared-seen.md)** — "seen" is one shared, Collie-only fact. It records the
two options that will otherwise be re-proposed forever: making it per-device (rejected — an alert
cleared on the phone must not still shout on the laptop), and counting a Herdr focus at the desk
(rejected — a pane clicked past would silently clear an alert you never read; a false positive costs
one tap, a false negative costs a missed agent).

## How it was checked

Built against a real 45-space herd, then refined over five evaluate → fix → re-capture rounds with
screenshots at 390 / 768 / 1280 and an objective audit (horizontal overflow, sub-36px tap targets,
heading tree, page height). That loop caught, among others: the title truncation above; section
headings rendering at two different sizes because a `<button>` doesn't inherit `text-transform` from
its `<h2>`; a hollow status ring that read as a notch cut out of the avatar; and — not from this
work — **light `--accent` being byte-identical to `--background`**, so every "this is the current
one" fill and every `hover:bg-accent` rendered as nothing in light mode.

- **351** bridge tests, **1170** web tests, both typecheck clean, `scripts/check-version.sh ✓`
- Verified live against the real herd: cold start seeds every pane as already-seen (a fresh install
  must not open on 45 unread alerts); reading a pane marks only that pane; the polling dashboard
  cannot false-positive, because `rootLoader` only ever fetches `/api/snapshot`

## Version

**0.19.0** — MINOR: additive, no config or API break.


AltanS:
Superseded by #47, now merged — thank you, this was excellent work.

Your 15 commits are carried verbatim and are in `main` with your authorship intact. The design is the part worth calling out: deriving the two timestamps locally because Herdr has none, and then making "unseen" a **comparison rather than a stored flag**, means there is no read-receipt state to keep in sync and no sync bug to write later. The backward-compatibility story falls out of stable sort and zero-returning comparators with no feature detection at all — that is the kind of thing most implementations get wrong by adding a `hasTimestamps` branch. And catching the read-level-GET CSRF hole in your own security review, before it shipped, is exactly the standard this repo wants.

Fixed on top:

- **A space row and its chip disagreed about colour** — the invariant the PR set out to enforce. `SpaceOverview` still ranked by `STATUS_RANK` while `SpaceStrip` had moved to `bucketOf`, so a space holding one working agent and one unseen-done agent read "working" on the dashboard and "ready" in the strip. Both share the classifier now, through a single-pass map that also removes an O(spaces × agents)-per-render cost.
- **The seen-stamp ran before the method dispatch**, so a request that 405s still cleared an alert.
- `paneTitle`, `paneSearchText` and `spaceLastSeen` had no production caller — their tests were the only thing keeping them alive. The behaviours still ship, so the tests moved to `paneParts` rather than being dropped.
- `paneTitleInTab` reimplemented `paneDisplayName` line for line; `aria-controls` named an unmounted id while collapsed; a `StatusDot` size override only resized its wrapper.
- Some UI polish found on-device: the Settings page shifted ~470px a frame after opening, divided rows had a radius with no border to own it, and the mirror had no top edge so the pane row bled into terminal output.

It ships in the untagged 0.20.0 alongside the theming work rather than taking its own version.

Issue #31, opened 2026-07-26
An installed PWA has no reachable path to a fronting proxy's sign-in page


Optic00:
I run Variant C behind a proxy that authenticates devices before anything reaches the bridge. Two
things combine so that a lost session leaves the installed PWA with nothing usable on screen and no
way back in. I hit this in my own deployment and had to build a workaround for it, which is really
the point of this report: the workaround is ugly in a way that suggests the shape upstream is
missing.

### The service worker answers every navigation from the precache

```ts
registerRoute(new NavigationRoute(createHandlerBoundToURL("/index.html"), { denylist: [/^\/api\//] }));
```

Every navigation except one under `/api/` is served from the precached app shell without touching
the network. Confirmed in the built bundle, not just the source: `web/dist/sw.js` on 0.14.2 contains
`denylist:[/^\/api\//]`.

So a sign-in or enrolment page served by the proxy at `/` is invisible to an installed PWA. It loads
the cached Collie UI instead, and there is no address bar to fall back to.

### The workaround this forces

My proxy serves its enrolment page at **`/api/__enrol`**. Not because that is a sensible place for a
page, but because the `/api/` denylist is the only crack through the precache. To be reachable from
an installed PWA, an operator has to put a non-API HTML page inside the namespace the API owns, and
order it ahead of the rule that would otherwise answer `/api/*` with a 401.

That works, and it is squatting. If Collie ever routed `/api/__*` itself, or tightened the denylist,
it would break silently. A supported escape hatch would remove the guesswork: a documented path that
the service worker always passes to the network, so a fronting proxy has somewhere legitimate to put
a sign-in page.

### The other half, and why it makes this worse

Until the operator can reach that page, they also cannot tell they need to. Every loader error
except `AbortError` lands in the same catch and returns the last-good data flagged stale, which the
connection banner renders as "reconnecting". A 401 from the proxy therefore looks exactly like a
dead radio. It is self-sustaining too: the escalation latch clears only via `markLive()`, so under a
persistent 401 it never clears.

That half is not proxy-specific. The bridge answers 403 from `checkAccess` and `guard` for
`host not allowed`, `cross-origin rejected` and `device not authorised`, so a misconfigured
`COLLIE_PUBLIC_HOSTS` or a device that is not on the allowlist presents the same way.

I have opened #30 for that half separately, since it seemed clear cut: classify 401 and 403 in the
loaders and let the banner say it is not a connection problem. `ApiError` already carries the status,
it was simply never read.

[...]

PR #43, opened 2026-07-28
fix: give a fronting proxy a reachable path from an installed PWA

AltanS:
Closes the routing half of #31 — the half left open when #30 shipped in 0.15.0.

## The problem

```ts
registerRoute(new NavigationRoute(createHandlerBoundToURL("/index.html"), { denylist: [/^\/api\//] }))
```

Every navigation except `/api/` was answered from the precached app shell without touching the
network. In an installed PWA there is no address bar to work around that, so a proxy's sign-in page
was unreachable wherever it lived — including at `/`, and including after a reload. @Optic00's
deployment worked around it by serving an enrolment page at `/api/__enrol`, because the API's own
namespace was the only crack in the precache. That is squatting a path Collie could reclaim at any
time, which is what they filed.

## What changes

`/auth/` is reserved. The service worker always passes it to the network, Collie routes nothing
there, and it will never become a UI route. The rules live in `web/src/lib/sw-routes.ts`, imported
by both the service worker and the app so the two can't drift on what the path is.

The bridge answers `/auth/*` itself with a plain-HTML placeholder saying nothing is configured
there, rather than letting the SPA fallback return the app shell. Without that, an operator whose
proxy rule is missing gets handed back the very UI they were trying to escape, with no signal at
all. The placeholder sits outside every gate — the reason to be there is that a gate refused you —
and carries a link home, because in an installed app it may be the only thing on screen.

The refusal banner from #30 now links to it. Its comment claimed "Reload is the useful action either
way, because it is what lets a fronting proxy serve its sign-in page" — the precache made that
false. Reload stays for a merely stale session on an already-enrolled device, but the escape is an
`<a href="/auth/">`: a real navigation the service worker sees as one, where a button's click
handler is a same-document action it never sees at all. There's a test pinning it as a link
specifically, since swapping in a Button would silently re-brick the PWA and nothing else would
fail.

## Checked

- The denylist survives minification into the shipped artifact: `grep` on built `dist/sw.js` gives
  `denylist:[/^\/api\//,/^\/auth(?:\/|$)/]`
- Live against the running bridge: `/auth/` → 404 placeholder, `/auth/sign-in` → same, `/` → 200,
  and `/authors` → 200 app shell (a prefix match must not be handed to the network)
- 366 backend tests, 1081 web tests, both typechecks clean, `scripts/check-version.sh` ✓

**Not verified here:** the installed-PWA path itself. Service workers need a secure context, so
confirming that a real installed app escapes the precache needs HTTPS and a device. The mechanism is
pinned statically and by unit test; the on-device behaviour is not.

## Version

**0.18.0** — MINOR, and it also cuts the release for #29, which merged unbumped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to Collie are recorded here. The format follows
 `version` in `herdr-plugin.toml`, `package.json`, and `web/package.json` (enforced by
 `scripts/check-version.sh`). See [`CLAUDE.md`](./CLAUDE.md) → *Versioning* for the bump policy.
 
+## [0.18.0] - 2026-07-28
+
+### Added
+- **Approvals are bound server-side to the prompt they were decided against.** `/keys` and `/reply` accept an optional `expected_prompt`; the bridge re-reads the pane immediately before writing and refuses with `409 prompt_changed` if the dialog moved. Shrinks the guard window from human latency to two local RPCs — a mitigation, not a guarantee, until herdr gains a conditional-input primitive (#29) — thanks @Optic00 (6afaf5b)
+- **`/auth/` is reserved for a fronting proxy's sign-in page**, and the service worker always passes it to the network. An installed PWA could not reach a proxy page at all before — the precache answered every navigation, reload included — so operators had to squat a page inside `/api/`. The refusal banner now links there (#31) — thanks @Optic00 (1a5972b)
+
 ## [0.17.0] - 2026-07-27
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -547,6 +547,33 @@ indefinitely — clients keep running old code with no way to notice. If your pr
 honor origin headers (Caddy and stock Nginx `proxy_cache` do by default; CDNs often need it
 enabled explicitly).
 
+**Serve your sign-in page under `/auth/`.** Collie reserves that path for you and routes nothing
+there. It matters because of how an installed PWA behaves: the service worker answers every
+navigation it owns from the precached app shell without touching the network, and there is no
+address bar to work around it. So a proxy page served anywhere Collie owns — including `/` — is
+invisible to the installed app, and a reload just re-renders the refused UI. `/auth/` (and anything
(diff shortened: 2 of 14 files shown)


AltanS:
Ran this end to end against a real Variant C topology before merging — a proxy that owns `/auth/`, 401s the API, and 302s navigations, in front of a bridge on the demo herd, driven through a real Chrome with the PWA actually installed.

**The bug, reproduced on the pre-fix build.** With the old service worker the origin is sealed harder than the issue describes: not only does Reload re-render the refused UI, but navigating *directly* to `/auth/` — the equivalent of typing it in an address bar — is answered from the precache and lands on Collie's error boundary. Nothing but `/api/` reaches the network.

**The fix, verified on the same origin.** Banner shows Sign in → tap → the proxy's page renders → sign in → back in Collie with the banner gone. Offline deep links to `/pane/:id` and `/settings` still resolve from the precache.

**One real bug found in review and fixed here (aa7765e).** Workbox matches the denylist against `url.pathname + url.search`, not the pathname — confirmed in the vendored `workbox-routing/NavigationRoute.js`. `/^\/auth(?:\/|$)/` therefore missed `/auth?rd=%2F`, the redirect shape Authelia and oauth2-proxy both emit, and the precache swallowed it. Every test passed because every test fed a pathname with no query string. Now `[/?]`, with tests that carry the query, and `/auth?rd=%2Fpane%2Fw1%3Ap1` confirmed reaching the proxy in the browser.

Also reserved `/cdn-cgi/`, since Cloudflare Access can't be moved off it and `/auth/` can't help there.

**One limitation worth knowing, measured not assumed.** A client already locked out cannot receive this fix: with the proxy refusing the bundle, `/sw.js` is 401 and `registration.update()` throws outright, so the old worker is permanent. Allowing the static bundle through while still 401-ing `/api/` fixes it — the worker then updates cleanly mid-refusal. Both states measured; the README now says so, along with the fact that a device that lapsed before 0.18.0 needs its site data cleared once.

The `/auth` reservation is implemented twice, so there's now one test that runs a shared case list through both the SW rule and the bridge's `isReservedAuthPath` and requires them to agree.

PR #45, opened 2026-07-29
feat(web): make URLs in the pane mirror tappable

konpyl:
URLs an agent prints into a pane are currently just text — you read them off the phone screen and
retype them. This makes them tappable.

### What it does

`http(s)://` runs in the mirror become anchors that open in a new tab. In practice this is mostly
agent tool output: `Fetch(https://grammy.dev/advanced/flood)`, a PR link, a localhost URL from a
dev-server log.

### How it works

`web/src/lib/links.ts` scans the visible mirror text and returns `[start, end, href)` ranges in the
**same coordinate space `find.ts` already uses**, so one running offset serves both splits. The
renderer applies links outermost and find matches inside, which means a search hit *inside* a URL
still highlights, and a URL that straddles an SGR colour change becomes one anchor per segment slice,
each carrying the same href.

That let the two render paths (fast + find-highlight) collapse into one. With no query and no links
the renderer still emits the segment's own string, so the polling path allocates exactly what it did
before.

### Safety

Only `http(s)` is matched, and the href is the matched text itself. `javascript:` and `data:` aren't
filtered out — they're **unmatchable**, which felt like the right shape for something rendering
untrusted agent output. A stray control byte can't reach an href either. Text still reaches the DOM
only as React text nodes; no `dangerouslySetInnerHTML` was added.

### Details worth a look

- **Trailing punctuation** is trimmed with paren balance respected: `See https://x.dev/a.` drops the
  full stop, `(https://x.dev/a)` drops the paren, `https://x.dev/Foo_(bar)` keeps its own.
- **Tap target.** At 12px × 1.25 leading an anchor is 14px tall, which is fiddly on a phone, so the
  anchor carries `py-[0.35em]`. Vertical padding on an *inline* box doesn't grow the line box, so
  this is free — I measured the `<pre>` at identical height with and without it, terminal grid
  untouched — while the target goes to ~22px. It has to stay em-relative: the pad is invisible, so a
  px value tuned for 12px text would reach past the neighbouring line's centre at the 9px floor of
  the A−/A+ control, and a tap on ordinary output would silently open a link. Verified in Chrome at
  9/12/16px that it doesn't.
- **Hard-wrapped URLs** link only as their first fragment. Stitching soft line breaks back together
  needs the pane's column width and a guess about which breaks were soft; a half-URL that opens the
  right host seemed better than a wrong URL assembled from two lines.
- **Bare `www.` hosts are deliberately not linked** — terminal output is dense with dotted tokens
  (paths, versions, IPs) and a host-shaped heuristic turns those into text you can't select.
- **OSC 8 hyperlinks aren't handled**; `parseAnsi` skips OSC entirely and I haven't seen one come
  out of a herdr pane.

### Verification

- 12 unit tests for the scanner, 7 for the rendering (`ansi-output.test.tsx` is new — there wasn't
  one).
- Full suite green on this branch: 1102 frontend, 440 backend. Build and `check-version.sh` clean.
- Swept in a real browser against live panes at 390×844, 768×1024 and 1440×900, both themes: correct
  hrefs, `rel="noopener noreferrer"`, no horizontal overflow, role `link` with an accessible name in
  the a11y tree, keyboard-focusable, tap opens the new tab and does not steal focus into the composer.

### Two things to decide

1. **The version bump is a guess** — I bumped 0.19.0 → 0.20.0 with a CHANGELOG entry, but you know
   your release timing. Re-cut or drop that commit as you like; the CHANGELOG entry has no commit
   hashes for the same reason.
2. **The underline colour.** It's `currentColor`, so the link inherits whatever colour the agent
   printed and stays legible on either theme. If you'd rather links read as one consistent colour,
   that's a one-line change.

Written against `main` at de8196b.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to Collie are recorded here. The format follows
 `version` in `herdr-plugin.toml`, `package.json`, and `web/package.json` (enforced by
 `scripts/check-version.sh`). See [`CLAUDE.md`](./CLAUDE.md) → *Versioning* for the bump policy.
 
+## [0.20.0] - 2026-07-29
+
+### Added
+- **URLs in the pane mirror are tappable** — `http(s)://` text becomes a link that opens in a new tab, keeping the colour the agent printed and marked by an underline
+- Trailing prose punctuation is trimmed, paren balance respected — `Fetch(https://x.dev/a)` links the URL, not the paren
+- A find hit inside a URL still highlights; a URL that changes colour mid-way stays one link
+
+### Changed
+- Link tap targets are `0.35em` taller than the text without moving the terminal grid — 14px is too small to hit on a phone
+
+### Fixed
+- Nothing but `http(s)` can become a link — `javascript:`/`data:` are unmatchable, not filtered, and a stray control byte can't reach an href
+
 ## [0.19.0] - 2026-07-29
 
 ### Added
--- a/herdr-plugin.toml
+++ b/herdr-plugin.toml
@@ -1,6 +1,6 @@
 id = "herdr.collie"
 name = "Collie"
-version = "0.19.0"
+version = "0.20.0"
 min_herdr_version = "0.7.0"
 description = "Mobile web UI to monitor and reply to your agent herd, served over Tailscale"
 platforms = ["linux", "macos"]
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "collie",
-  "version": "0.19.0",
+  "version": "0.20.0",
   "private": true,
   "license": "MIT",
   "description": "Collie — a mobile web UI to monitor and reply to your Herdr agent herd over Tailscale",
--- a/web/package.json
+++ b/web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "collie-web",
-  "version": "0.19.0",
+  "version": "0.20.0",
   "private": true,
   "license": "MIT",
   "type": "module",
--- a/web/src/components/ansi-output.test.tsx
+++ b/web/src/components/ansi-output.test.tsx
@@ -1,0 +1,76 @@
+import { describe, expect, it } from "vitest";
+import { render } from "@testing-library/react";
+import type { ComponentProps } from "react";
+
+import { AnsiOutput } from "./ansi-output";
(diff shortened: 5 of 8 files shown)


AltanS:
Superseded by #48, now merged — thank you, another really nicely judged one.

Your commit is carried verbatim, so it's in `main` with your authorship. The part worth calling out is the security shape: making `javascript:`/`data:` **unmatchable rather than filtered**, and keeping `href` a prefix substring of the matched text so it can never diverge from the characters on screen. That's a stronger guarantee than any denylist, and it's the right instinct for something rendering untrusted agent output. Reusing `find.ts`'s coordinate space so one offset serves both splits is what made the two render paths collapsible at all — that's a good structural call, not just a tidy one.

Two things needed doing on integration, both because this branched before the theming work landed:

- **Your unified `renderFind` carried the pre-theming highlight styling.** The mirror now renders in dark space under every theme and light mode inverts the whole `&lt;pre&gt;` ([ADR 0002](https://github.com/AltanS/collie/blob/main/.adr/0002-invert-the-light-terminal-mirror.md)), which forced an asymmetry in the find highlight: the current match re-applies the filter to cancel it (yellow-400 otherwise inverts to a brown that reads as a redaction bar), and can only do that because `text-black` pins its text. A non-current match sets no colour, so double-inverting would send its inherited text light → dark → light and erase the hit. Restored inside your structure.
- `ansi-output.test.tsx` was an add/add — there is one now, guarding those colour-space rules. Merged so the file has both suites; all 11 pass, which is the useful proof your refactor kept the theming behaviour.

One correction to a comment: the tap-target note said the padded box "stays inside the 1.25 line-height". It doesn't — ~14px content plus 2×4.2px padding is ~22px against a 15px advance, as the numbers in the same comment imply. It overlaps by design, and what makes that safe is the two-part invariant you were actually relying on: the pad never reaches the neighbouring line's *centre* at any size the A+/A− control offers (which is exactly why it must stay em-relative), and where it does overlap real text, that line's spans win inline hit-testing by DOM order. Left as written, someone would eventually "fix" the padding to match and delete the tap target.

On your two questions: kept `currentColor` for the underline — a fixed colour would fight the light-mode inversion — and dropped the version bump, since this ships in the untagged 0.20.0 with the theming and dashboard work.

PR #49, opened 2026-07-29
feat(web): consolidate terminal toolbar controls

simonallfrey:


AltanS:
Shipped the direction here in 0.23.0 (a0f880e) — thank you, this PR is what started it, and the changelog says so.

Different execution in the end: display prefs went behind a ⚙ into a labelled panel rather than a horizontally-scrolling toolbar (the raw-terminal `>_` needed a name more than it needed a smaller home), Keys/Quick/Agent kept their labels, and Find moved to the header where its find bar already takes over the row. On top of that: press feedback on every key and quick reply, hold-to-repeat on the arrows, and haptics.

Closing as superseded rather than rejected.

PR #51, opened 2026-07-29
Read two AskUserQuestion shapes that fall through to the raw mirror

konpyl:
Two AskUserQuestion shapes currently fall through to the raw mirror. On a phone that means a wall of ~180-column terminal text with nothing to tap and no way to answer — both were hit in normal use today, in different repos.

Both come from the same root mistake: a grammar **asserting** a layout instead of reading it.

## 1. A wrapped option label kills the preview dialog

`preview-select` required its numbered rows to be on consecutive lines. The assumption is written down as fact:

```
// 3. Option rows: numbered rows 1..k in CONSECUTIVE lines (the left column
//    never wraps its 30-col labels)
```

That gutter is only ~30 columns, so Claude Code wraps any longer label onto continuation rows and the numbered rows stop being adjacent — the whole dialog bails to raw. Observed live: an option labelled `Rename + keep old as hidden aliases (Recommended)` occupied three rows, putting the options on lines 183/186/187.

Reproduced by running the real `parseAnsi → splitLines → buildBlocks` pipeline over the capture, and flipped by the single change of making those rows adjacent — nothing else about the dialog mattered.

Adjacency is replaced by a contiguity walk keyed on the **label column** (the width of the `❯ 1. ` prefix): left of it is a new option, at it is label text, right of it is not label text at all. That one rule also prevents a label ending `…and 3. Backfill later` from minting a phantom third button, and stops preview-pane box-drawing being folded onto a button in layouts where the `Notes:` column and the preview pane don't coincide.

## 2. A multiSelect question inside a multi-question wizard is owned by nobody

`wizard` refuses it **correctly** — a wizard digit selects *and advances*, a checkbox digit only toggles, so claiming it would mis-drive the dialog — and `multi-select` only ever handled the single-question form. Two gates encoded that: a stepper of exactly two chips, and an advance row spelled `Submit`.

`multi-select` now carries the wizard's chips (via the wizard's own `parseStepperLine`, so the chip grammar stays in one place) and navigates with the wizard's own Left/Right keys, following the pattern `preview-select` already established for this exact shape. The advance row's label is **read, not assumed**: it is `Next` on every step but the last.

The row is located by **position** — last non-blank line above the rule separating the menu from the escape row — never by scanning for the word. Pane text is model-authored; matching on text alone let an option *description* reading `Next` rename the button to an action it doesn't perform, satisfy the "row absent → fail closed" check on a pane with no advance row, and forge the pointer so `Enter` landed on whatever row the terminal's `❯` was really on (on `Chat about this`, that aborts the whole tool call).

The advance macro stops at the keystroke rather than predicting what follows — it verifies the pointer in the same iteration it presses Enter, then lets the next poll re-detect whatever appeared. The single-question review stays where it was; a multi-question dialog reviews once at the end, and that screen belongs to `wizard`.

## Evidence

The `claude--wizard-multiselect-*` fixtures are sandbox captures of a real two-question dialog driven end to end, because nothing in the repo described this shape and guessing feeds a keystroke into a live terminal. They settled four things worth knowing:

- the question chip flips `☐ → ☒` on the **first** tick — "answered" means touched, not complete;
- digits toggle pointer-independently inside a wizard step, exactly as standalone;
- the last step's row reads `Submit`, earlier ones `Next`;
- the footer is **not** invariant — it gains `ctrl+g to edit in Vim` when the pointer reaches the advance row, which is why the signature already stops before it.

`claude--wizard-preview-wrapped-label.txt` is **derived, not captured**: its live specimen was real work and this repo is public, so the gutter of `claude--wizard-preview-q1.txt` was rewritten and everything from the `Notes:` column rightward carried over untouched. Flagged as derived in the fixtures README.

Existing single-question dialogs are unchanged (`steps: null`, review phase untouched), and every non-dialog fixture still yields zero detections.

## Also in here

- The race guard now compares the wizard steps. `signature` normalises `☒/☑ → ☐` across the whole chip line — it has to, since the current chip flips on the first tick — but that also erases *which* step you are on, so two questions of one wizard sharing a question text and option labels were byte-identical to the guard.
- The stepper strip is one shared component; `wizard`, `preview-select` and `multi-select` all render it from the same place instead of three copies.
- `multi-select-action.ts` carried a literal NUL byte as a map-key separator, which made git classify the file as **binary** and hide every change to the race guard from `git diff`. It's now a source escape, plus a `.gitattributes` rule so a future stray byte can't cost reviewability.

## Notes for review

- **One commit, code only.** No version bump and no CHANGELOG entry — this comes from a fork with its own numbering lane, so slotting it into your release is yours to do.
- Four new fixtures are pinned in `prompt-binding-regions.json`; the completeness gate and the bridge-side binding test both pass.
- Gate on top of `main`: **1271** web tests pass, `tsc` clean.


konpyl:
Rebased onto `main` (`ae8f562`, 0.20.2) — it was cut from `c674af5`, so it had 33 commits of drift. Still one commit, no conflicts, and re-verified against the current tree: `tsc` clean, **1271** web tests and **479** bridge tests pass.

Worth confirming the change is still needed rather than just still applying: `preview-select.ts` on `main` today still carries the adjacency check —

```ts
if (rows.some((r, k) => r.n !== k + 1 || r.index !== rows[0]!.index + k)) return null;
```

— so a wrapped option label still drops the whole dialog to the raw mirror.

Also, thank you for `4841e37`. You were right, and it's a better fix than mine: `command -v` reporting a shell function as a bare word makes `dirname` return `.`, and my version would then have put the working directory on the `PATH` used by every later `git`/`systemctl`/`tailscale` in that script. I've cherry-picked it into my fork and reproduced both behaviours to confirm. The point about the lifecycle suite stubbing `BUN=/bin/true` — and that being exactly why a missing Bun failed four deploys unnoticed — landed too.

AltanS:
Landed via #62 with your commit cherry-picked verbatim (authorship preserved, bdf4c26) — thanks @konpyl, this ships in the next release.

PR #52, opened 2026-07-29
fix(scripts): find Bun when Herdr invokes a plugin action

konpyl:
`herdr plugin action invoke update` fails with `error: bun not found on PATH`.

Herdr spawns plugin actions with a minimal environment — no login shell, so nothing has sourced the line Bun's installer adds to your profile and `~/.bun/bin` isn't on `PATH`. `collie-ctl.sh:47` resolves Bun with `command -v bun` alone, so every action that needs it dies.

## Why this is worth fixing rather than documenting

It's not that it fails, it's *how*:

- The systemd unit is written with an **absolute** `ExecStart`, so the running bridge is unaffected. Only plugin-invoked actions break — `update`, `restart`, `start`.
- `update` **pulls before it builds**:

  ```sh
  cmd_update() {
    git -C "$PLUGIN_ROOT" pull --ff-only
    exec bash "${PLUGIN_ROOT}/scripts/collie-ctl.sh" _apply-update   # ← dies here
  }
  ```

  So a failed run still fast-forwards the checkout. You end up with new code on disk, the **old `web/dist` still being served**, and both the manifest and `/api/config` reporting the new version. A half-succeeded deploy is indistinguishable from a working one unless you go read the plugin log.

On my deployment it failed this way four times before I noticed — I only caught it because I checked the built bundle rather than the version string.

## The change

Resolve Bun from `PATH` first, then `$BUN_INSTALL`, `~/.bun/bin`, `~/.local/bin`, `/usr/local/bin`, `/opt/homebrew/bin`, `/usr/bin`.

Its directory is prepended to `PATH`, not just stored in `$BUN`: the script shells out to a bare `bun` for the Tailscale ownership probe (`collie-ctl.sh:94`), and `bun run build` spawns children that expect to find it themselves.

An empty result stays non-fatal — the existing `[ -n "$BUN" ] || { echo "error: bun not found on PATH"; exit 1; }` guards still report and exit, so a host genuinely without Bun gets the same message as before.

## Verification

- Reproduced the failing environment with `env -i HOME=$HOME PATH=/usr/bin:/bin`: resolves and puts Bun on `PATH`.
- `bash -n` clean; `bun run test` (474 tests incl. the `collie-ctl` lifecycle suite) passes on top of `main`.
- End to end on a real deployment: the same action that had been failing now completes. It repairs itself on the first run, since `cmd_update` re-execs the just-pulled script.

One commit, one file, no version bump or CHANGELOG entry — this comes from a fork with its own numbering lane, so slotting it into your release is yours to do.

### DIFF
--- a/scripts/collie-ctl.sh
+++ b/scripts/collie-ctl.sh
@@ -77,13 +77,19 @@ resolve_bun() {
 BUN="$(resolve_bun)"
 # Put it on PATH too, not just in $BUN: this script shells out to a bare `bun` (the Tailscale
 # ownership probe), and `bun run build` spawns children that expect to find it themselves.
-if [ -n "$BUN" ]; then
-  BUN_DIR="$(dirname "$BUN")"
-  case ":${PATH}:" in
-    *":${BUN_DIR}:"*) ;;
-    *) PATH="${BUN_DIR}:${PATH}"; export PATH ;;
-  esac
-fi
+#
+# ABSOLUTE paths only. `command -v` reports a shell function or alias as a bare word, and the plugin
+# .env is sourced above us — so a `bun()` defined there would resolve to `bun`, whose dirname is `.`,
+# and we'd prepend the CWD to the PATH used for every later `git` / `systemctl` / `tailscale`.
+case "$BUN" in
+  /*)
+    BUN_DIR="$(dirname "$BUN")"
+    case ":${PATH}:" in
+      *":${BUN_DIR}:"*) ;;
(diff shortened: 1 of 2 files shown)


AltanS:
Shipped in 0.20.2. Added the absolute-path guard on the PATH prepend (`command -v` reports a `bun()` from the .env as a bare word, so the CWD was reaching `git`/`systemctl`) and the resolution's first test coverage. The write-up is what made this quick to confirm — thank you.

Issue #34, opened 2026-07-26
PWA send renders in the pane's input box but never reaches the input buffer (silently lost)


maikschuheida-spec:
## Summary

Text sent from the Collie composer intermittently never reaches the harness: the sent text **renders inside the harness's input box in the terminal mirror**, but the harness itself never receives it - the message is silently lost (no send, no error surfaced in the PWA). Happened 6+ times in a single day across different panes.

## Environment

- Collie 0.14.0 (all occurrences); per changelog review, 0.14.1/0.14.2 contain no related fix, so presumably still present in 0.14.2 (now running it, no long-term data yet)
- Herdr 0.7.4
- Bridge host: macOS (Apple Silicon), bridge behind `tailscale serve` (README Variant A)
- Client: iOS PWA installed to the home screen
- Harness in the affected panes: Claude Code

## Steps to reproduce (as observed; intermittent)

1. Open a pane running Claude Code in the PWA.
2. Type a message in the composer and tap Send.
3. The terminal mirror shows the text rendered in the harness's input box - looks like a successful hand-off.
4. The harness never processes it: nothing is submitted, and checking the terminal directly shows the input buffer does not actually contain the text.

## Expected

Either the text lands in the harness input buffer (and submits), or Collie surfaces a failure so the user knows the send was lost.

## Actual

The mirror renders the text as if delivered, but the input buffer never received it. No error, no retry hint - the message silently disappears. From the phone this is indistinguishable from a successful send until you notice the agent never responded.

## Notes

- Intermittent: most sends work; on the affected day it failed 6+ times.
- Happy to pull bridge/audit logs for a failing occurrence on request.

AltanS:
**Reproduced, root-caused, and fixed in 0.17.0.** Thanks for this — it turned out to be more serious than the report, so I'll lead with that part.

**It isn't only losing your message. It's answering dialogs.**

I reproduced it on a real Claude pane in manual mode. With a Bash permission prompt focused and `1. Yes` highlighted, I sent a message through the same path the phone uses:

- the bridge returned `{"ok":true}`
- the dialog was dismissed and **the command ran**
- the message appeared nowhere

The submit key answered the prompt. Claude highlights `Yes` by default, so a reply typed at the wrong moment can approve a tool call you never saw. That's the part worth knowing about if you've been driving agents from your phone.

[...]

PR #66, opened 2026-08-06
fix(web): verify wrapped CJK drafts so the reply guard submits them

tyamanak:
## The bug

`draftCarriesSend` is the #34 guard: it withholds the submit key until the typed text is verifiably in the input box, by requiring the visible draft to be a contiguous slice of what we sent.

`extractInputDraft` folds a wrapped draft's visual lines together with a space. That restores a **real** space only when the box wrapped at a word boundary. CJK has no spaces to wrap at, so a mid-run wrap fabricates a space the send never had, the slice check can never match, and the guard withholds Enter forever:

```
sent    "これちなみに電池寿命的にはどうなんだろね。"
screen  ❯ これちなみに電池寿命的にはどうなん
          だろね。
draft   "これちなみに電池寿命的にはどうなん だろね。"   ← space not in the send
        → collapse(sent).includes(collapse(draft)) is false, always
```

Every wrapped Japanese reply stalls unsubmitted. Whether a given message wraps depends on the pane width, so it presents as flaky rather than deterministic — short messages go through, longer ones silently don't.

## Why not just ignore whitespace, or branch on language

Ignoring whitespace on both sides fixes the stall but over-corrects: it also accepts a screen genuinely differing from the send by a space, which is exactly what the guard exists to catch.

Branching on "is this CJK?" is the wrong axis. The question isn't what language the text is, it's **whether a given seam was a real space** — and one draft can hold both kinds at once:

```
sent    "これは pull request のテストです"
draft   "これは pull request のテ ストです"
                ↑ real           ↑ fabricated
```

The ambiguity is per-seam, not per-string, and after the join both look identical.

## The fix

Loosen exactly the seam and nothing else. Split the draft into runs and the gaps between them; a gap spelled exactly like the fold's own seam (one plain space) may collapse to nothing, and every other gap must appear in the send verbatim, since the fold cannot produce it. Runs still match character for character, in order and contiguously.

```
draft    "これは pull request のテ ストです"
runs     ["これは", "pull", "request", "のテ", "ストです"]
pattern  /これは\s*pull\s*request\s*のテ\s*ストです/
sent     "これは pull request のテストです"     ✓
```

The same `\s*` covers a real space and a zero-width seam, so nothing has to decide which is which.

Three supporting corrections keep the check honest at the character level:

- **Non-seam gaps match verbatim.** A screen holding `危険　実行` (U+3000) no longer verifies a send of `危険実行`. The tolerance is one-directional, keyed on what the draft shows, so wrapping *at* a full-width space still folds to the seam and still passes.
- **Matches land on grapheme boundaries.** A code-unit substring can cut a cluster in half — `👩‍👧‍👦` sits inside `👨‍👩‍👧‍👦` without being it. Both ends are checked, and every occurrence is tried rather than only the first, so an earlier misaligned hit can't mask a later valid one.
- **The floor counts readable clusters,** not UTF-16 units. It segments the string as written: stripping spaces first would fuse neighbours (`🇯 🇵` into one flag) and halve the floor. Invisible formatting controls (LRM, ZWSP, soft hyphen) no longer pad it.

## Safety

The guard's protection is unchanged. A focused permission dialog leaves no input box on screen, so `extractInputDraft` returns `null` and no key is sent — verified against all of `web/src/fixtures/panes/`: every dialog capture (`permission-*`, `plan-approval*`, `trust-prompt`, `select-*`, `wizard-*`) returns `null`.

What was loosened is only the width of a gap the fold itself could have produced. What was tightened — verbatim non-seam gaps, grapheme-aligned matches, an honest floor — all runs in the false-positive direction, which is the dangerous one.

## Tests

Ten cases added, with as much weight on what must still be **rejected** as on what should now pass: wrapped CJK, mixed-script drafts at both kinds of seam, windowed tails, dropped or altered characters, non-contiguous runs, full-width vs half-width gaps, grapheme boundaries at both ends, multiple occurrences, invisible-control padding, and regex metacharacters treated as literals.

Full suite green (`bun run test` in `web/`, plus root `bun run test`); both sides typecheck.

### DIFF
--- a/web/src/lib/reply-action.test.ts
+++ b/web/src/lib/reply-action.test.ts
@@ -57,6 +57,106 @@ describe("draftCarriesSend", () => {
   it("rejects an unrelated draft", () => {
     expect(draftCarriesSend("deploy to prod", "someone else's leftover")).toBe(false);
   });
+
+  it("accepts a CJK draft wrapped mid-run (the fold fabricates a space the send never had)", () => {
+    // A Japanese draft has no word boundaries, so the input box wraps it mid-run and the
+    // space-joined fold yields a space absent from the sent text. This stalled every wrapped
+    // Japanese reply: the guard never verified the text and withheld the submit key.
+    const sent = "これちなみに電池寿命的にはどうなんだろね。";
+    expect(draftCarriesSend(sent, "これちなみに電池寿命的にはどうなん だろね。")).toBe(true);
+    // A windowed (tail-only) slice of a wrapped CJK draft still matches.
+    expect(draftCarriesSend(sent, "電池寿命的にはどうなん だろね。")).toBe(true);
+    // An unrelated CJK remnant still fails.
+    expect(draftCarriesSend(sent, "別の誰かの下書きです、これは。")).toBe(false);
+  });
+
+  it("accepts mixed CJK/latin text wrapped at either kind of seam", () => {
+    // The case no language test could handle: ONE draft carrying both a genuine space (between
+    // "pull" and "request") and a fabricated one (wherever the box broke the CJK run).
+    const sent = "これは pull request のテストです";
+    expect(draftCarriesSend(sent, "これは pull request のテ ストです")).toBe(true); // mid-CJK break
+    expect(draftCarriesSend(sent, "これは pull request のテストです")).toBe(true); // at the space
+    expect(draftCarriesSend(sent, "これは pull request のテストです")).toBe(true); // no wrap at all
+  });
+
+  it("still rejects a draft that lost or altered a non-space character", () => {
+    // Only the WIDTH of a gap is unknowable — every visible character must still be there. A box
+    // showing text with a space genuinely missing is NOT our text and must not be verified.
+    expect(draftCarriesSend("deploy the app", "deploythe app")).toBe(false);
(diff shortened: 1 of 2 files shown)


AltanS:
Thanks!! Merged and shipped in [0.24.2](https://github.com/AltanS/collie/releases/tag/v0.24.2) — reproduced on a live pane first (the mid-run wrap is exactly as you described), and the analysis in the PR body made the review easy; thank you. One follow-up landed on top (46a85d1): the module-scope `Intl.Segmenter` is feature-detected now, so engines without it (Firefox < 125 / Safari < 14.1) fall back to code points instead of failing to boot.

AltanS:
Heads-up: this commit was authored as `tyamanak@Kotori-MBP.local`, which GitHub can't link to your account — so it won't count on the contributors page. Setting `git config user.email` to your GitHub (or noreply) address will make future PRs count.

tyamanak:
Thanks for merging it! And sorry for missing the browser-compatibility angle in my own review — thank you for taking care of it.

Thanks for the note about the author address too. I switched machines recently and forgot to set it up again.

I really appreciate you building such a great project. I'll keep using it, and I'll send a PR whenever I notice something.

PR #71, opened 2026-08-07
feat(windows): add native plugin lifecycle

JJLiebig:
Completes the native launcher gap left after #27 added Windows named-pipe transport. Windows can now install and run Collie through the same Herdr actions as Linux and macOS.

## What changes

- Adds a PowerShell lifecycle controller backed by Task Scheduler.
- Generates one immutable platform-native action launcher used by every host.
- Updates both Herdr-managed detached installs and linked clones without changing their install type.
- Keeps Tailscale Serve operator-managed on Windows; Collie only reports the existing ingress.
- Keeps full application CI on Linux and adds Windows lifecycle coverage on standard GitHub-hosted runners.

## Defaults

- The Windows task runs with limited privileges by default.
- If Herdr intentionally runs as Administrator, COLLIE_TASK_RUN_LEVEL=highest is an explicit opt-in; phone actions then inherit Administrator power.
- The bridge remains bound to loopback behind Tailscale Serve or another operator-owned ingress.
- Linux and macOS keep their existing supervisor and managed-Serve behavior.

## Verification

- Linux bridge suite: 534 passed; POSIX lifecycle passed before the final launcher correction, with its changed launcher generation rechecked under WSL.
- Windows lifecycle suite: passed, including canonical \\?\ paths and both task run levels.
- Web suite: 1,603 passed.
- Root and web typechecks: passed.
- Herdr invoked version and restart successfully from the linked canonical path.
- Windows production build and live HTTPS bridge: 0.26.0+8192f01, connected with 20 agents.
- Review Suite fast: green (rvw_6f575a57).

## Notes

This really isn't a sort of "merge as is" more of a reference how I got this to work quite nicely on Windows.

I'm still minimizing it but figured maybe it helps serve as reference.

Edit: Did a fair bit of minimizing, can now be considered reasonable to merge.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,11 +21,11 @@ COLLIE_HOST=127.0.0.1
 # Tunnel: Collie publishes nothing, you point your tunnel at 127.0.0.1:$COLLIE_PORT and start it
 # however you start your other services (README → Variant E).
 # COLLIE_SKIP_SERVE=1
-# Your public URL behind the proxy — shown as the "proxy" address by `collie-ctl.sh status`.
+# Your public URL behind the proxy — shown as the "proxy" address by the status action.
 # COLLIE_PUBLIC_URL=https://collie.example.com
 
 # --- Herdr connection ---
-# Usually injected by the systemd unit; override only if your socket is elsewhere.
+# Usually injected by the native supervisor; override only if your socket is elsewhere.
 # Defaults to ~/.config/herdr/herdr.sock, or %APPDATA%\herdr\herdr.sock on Windows.
 # HERDR_SOCKET_PATH=/home/you/.config/herdr/herdr.sock
 # Which dialer opens that socket: auto (default) | net | bun. "auto" is right everywhere — node:net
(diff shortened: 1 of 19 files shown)


AltanS:
Thanks, this is good work and I don't want to lose it. But I can't take Windows on as a supported platform: every lifecycle change would land twice, tested on an OS I don't run.

Middle path: your controller, launcher and tests now live in `contrib/windows/` (8572e49), your authorship preserved, README distilled from this PR. It's community-maintained: works as of 0.27.x, outside CI and the release gate, fixes come as PRs. Consider it yours.

Left out on purpose: version bump, CHANGELOG, CLAUDE.md, and the check-version rewrite (a release-gate change I'd want to make deliberately, not alongside a feature). The manifest also stays POSIX-only, so the lifecycle is driven from PowerShell; the README covers pointing a local manifest edit at your launcher if you want the buttons.

Closing in favor of that commit. If Windows interest grows, promoting contrib/ to supported is on the table. Thanks again.

PR #73, opened 2026-08-08
feat(harness): add Pi read-only adapter

en-ver:
## Summary
- Add a read-only Pi harness adapter for the verified stock Pi 0.82.1 inline editor only.
- Recognition is intentionally narrow and fail-closed: overlays and custom editors stay in the raw terminal mirror rather than becoming writable composer state.
- Reply verification is baseline-aware for normal typing and paste paths, requiring an observed changed draft before submission.
- Add real, sanitized Pi pane fixtures with documented provenance.

## Release / compatibility
- No dependencies, lockfile changes, migrations, or configuration changes.

## Validation
- `bash scripts/check-version.sh`
- `bun install --frozen-lockfile && bun run typecheck && bun run test` — 534 tests passed
- `cd web && bun install --frozen-lockfile && NODE_OPTIONS=--no-experimental-webstorage bun run typecheck && NODE_OPTIONS=--no-experimental-webstorage bun run test` — 1,619 tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to Collie are recorded here. The format follows
 `version` in `herdr-plugin.toml`, `package.json`, and `web/package.json` (enforced by
 `scripts/check-version.sh`). See [`CLAUDE.md`](./CLAUDE.md) → *Versioning* for the bump policy.
 
+## [0.26.0] - 2026-08-08
+
+### Added
+
+- **Pi gets a read-only adapter** — its verified stock inline editor yields a draft while unrecognised terminal state stays raw, and reply submission requires a changed observed draft (2efea2f)
+
 ## [0.25.0] - 2026-08-07
 
 ### Added
--- a/HARNESS_CONTRIBUTING.md
+++ b/HARNESS_CONTRIBUTING.md
@@ -2,8 +2,8 @@
 
 Collie up-levels an agent's terminal dialogs (permission prompts, AskUserQuestion menus, plan
 approvals, …) into native phone buttons. The per-agent knowledge that makes this safe lives in a
-**harness adapter**. Claude Code is the one verified adapter today; this is how you add another
-(codex, pi, opencode, …).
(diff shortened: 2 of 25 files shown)


en-ver:
Closing because local visual acceptance failed. This contribution will remain closed pending correction and validation.

aspiers:
Heads-up on an overlap, and a note on how I'm sequencing it.

I've been building a **Tier 2 pi adapter for `ask_user` dialogs** — the questionnaire overlay from the `pi_ask_user` extension lifted into native phone buttons, both `displayMode`s (`overlay` and `inline`), single-select and multiSelect. It's verified against both my fork and real upstream `edlsh/pi-ask-user@main`, with the keystroke recipe cited line-by-line to pi's own handlers and a `WIZARD_NOTES`-style choreography file.

Our scopes look complementary rather than duplicated — yours is the composer (editor draft, paste, `composerReady`), mine is dialog grammars — but we both touch `harness/pi/index.ts`, `types.ts`, `registry.ts` and `conformance.test.ts`, so they'll conflict textually.

So you know what's coming: **I'm rebasing mine on top of this branch now**, so you're not resolving a three-way merge and the stacking order is obvious. If #73 gets force-pushed or restructured in the meantime, no problem at all — I'll just redo the rebase; please don't hold anything up on my account. Equally happy to rebase onto `main` after this lands instead, if that suits you better.

@AltanS — flagging in case you'd rather see these as one PR, or in a particular order.

en-ver:
Following further testing, I decided not to pursue the Pi-specific harness-adapter approach in this PR.

The underlying goal was to avoid messy output on small screens when terminal-width horizontal border and separator rows around the input area wrapped into multiple visual rows. I addressed that more narrowly in [#79](https://github.com/AltanS/collie/pull/79) by visually clipping long standalone terminal rule rows to the mirror width in wrap-on mode, while preserving their full text, ordinary output wrapping, and wrap-off horizontal panning.

[#79](https://github.com/AltanS/collie/pull/79) supersedes this PR.

PR #77, opened 2026-08-09
feat(composer): add Immediate terminal input mode

### LINKED ISSUE #74: Composer: add a mode for sending key sequences without Enter
## Summary

Collie has no way to send a bare printable key such as `b` from the phone UI.

The reply composer always submits the text with Enter. The Keys dock can send
special keys, digits, and modifier chords, but it cannot send an unmodified
letter.

## Requested behaviour

Add a user-selectable send mode to the composer:

- **Reply mode:** preserve the current behaviour—send the text and submit it.
- **Keys mode:** send the entered characters as a key sequence without appending
  Enter.

For example, entering `b` in Keys mode should send the equivalent of:

```json
{ "keys": ["b"] }
```

Longer input could be sent as an ordered sequence of literal character keys.
The existing Reply mode should remain the default.

## Why

Some terminal interfaces use bare letters as commands or menu shortcuts. These
cannot currently be activated through Collie unless Collie provides a dedicated
button for that particular key.

## Existing support

This appears to be a UI limitation rather than a Herdr limitation:

- `pane.send_keys` accepts literal one-character keys.
- Collie's `/keys` endpoint already accepts an ordered key array.
- The composer already has a separate non-submitting transport internally via
  `submit:false`, although this request specifically concerns sending key
  sequences.

Related but distinct: #19 and #20 added modifier and chord support to the Keys
dock; neither provides bare printable letters.

aspiers:
Fixes #74.

## The gap

The composer always followed text with the harness submit key. The Keys dock could send digits,
special keys and chords, but the phone keyboard could not drive a TUI that expects bare printable
keys such as `b`.

## Immediate mode

Hold the normal Send button to arm **Immediate mode**. The textarea and button change visibly, and
committed phone-keyboard input goes straight to `pane.send_keys` instead of becoming a reply:

- ordinary text is case-preserving and ordered;
- Space, Tab and explicit line breaks use Herdr's verified key names;
- Android/physical Backspace and Enter are captured even when they do not change the textarea;
- no trailing Enter is added;
- a tap on the highlighted keyboard button exits.

The mode survives ordinary blur/keyboard dismissal, but never a pane switch. It refuses to arm over
a buffered reply, and reply-only actions (Take over, slash-command insertion, image upload) disarm
it rather than reinterpret their text as terminal keys.

## Ordering and failure

Herdr uses one connection per RPC, so concurrent `send_keys` calls have no ordering guarantee. A
one-in-flight pump sends the first key immediately, accumulates later input behind it, and flushes
bounded 64-key batches in order. A slow tailnet therefore grows the next batch rather than reordering
characters.

If a batch fails, queued keys are discarded, Immediate mode stops, and the textarea blurs so
continued typing cannot silently become a normal Reply with an eventual Enter attached.

## Verification

- `bun x tsc --noEmit` — root typecheck clean
- `bun run test` — 534 bridge tests plus the collie-ctl lifecycle suite
- `cd web && bun run typecheck` — client and service-worker typechecks clean
- `cd web && bun run test` — 1,616 tests passed
- `cd web && bun run build` — production PWA build passed
- `scripts/check-version.sh` — 0.26.0 consistent

The tests exercise the Android `contextmenu` long-press fallback, native `beforeinput` Backspace,
IME-style committed strings, ordering/backpressure, blur persistence, pane reset, and failure stop.
A real Android/Gboard smoke test is still needed; jsdom cannot reproduce an OEM keyboard's exact IME
event stream.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to Collie are recorded here. The format follows
 `version` in `herdr-plugin.toml`, `package.json`, and `web/package.json` (enforced by
 `scripts/check-version.sh`). See [`CLAUDE.md`](./CLAUDE.md) → *Versioning* for the bump policy.
 
+## [0.26.0] - 2026-08-08
+
+### Added
+
+- **Hold Send to type straight into the terminal** — Immediate mode routes the phone keyboard through ordered key batches with no appended Enter, stays armed through keyboard dismissal, resets on a pane switch, and stops on a transport failure (#74, 4f5a253)
+
 ## [0.25.0] - 2026-08-07
 
 ### Added
--- a/herdr-plugin.toml
+++ b/herdr-plugin.toml
@@ -1,6 +1,6 @@
 id = "herdr.collie"
 name = "Collie"
-version = "0.25.0"
+version = "0.26.0"
 min_herdr_version = "0.7.0"
 description = "Mobile web UI to monitor and reply to your agent herd, served over Tailscale"
 platforms = ["linux", "macos"]
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "collie",
-  "version": "0.25.0",
+  "version": "0.26.0",
   "private": true,
   "license": "MIT",
   "description": "Collie — a mobile web UI to monitor and reply to your Herdr agent herd over Tailscale",
--- a/web/package.json
+++ b/web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "collie-web",
-  "version": "0.25.0",
+  "version": "0.26.0",
   "private": true,
   "license": "MIT",
   "type": "module",
--- a/web/src/components/composer.test.tsx
+++ b/web/src/components/composer.test.tsx
@@ -317,6 +317,39 @@ describe("Composer — Immediate key mode", () => {
     expect(screen.getByTestId("status")).toHaveTextContent(/immediate mode on/i);
   });
 
+  it("sends a swiped/IME-composed word once when composition commits", async () => {
+    const keyCalls: string[][] = [];
+    server.use(
+      http.post(/\/api\/pane\/[^/]+\/keys$/, async ({ request }) => {
+        keyCalls.push(((await request.json()) as { keys: string[] }).keys);
+        return HttpResponse.json({ ok: true });
+      }),
+    );
+    renderComposer();
+    const box = activateImmediateMode();
+
+    fireEvent.compositionStart(box);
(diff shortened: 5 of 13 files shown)


AltanS:
Shipped in #84, released as [v0.26.0](https://github.com/AltanS/collie/releases/tag/v0.26.0). Your commit is in `main` with authorship preserved (0acdc80) — thank you, this was a lot of careful work and the hard parts of it went in untouched.

Kept as submitted: the one-in-flight pump with ordered, bounded batches (the right call — Herdr opens a connection per RPC, so `hello` really can arrive as `hlelo`), the Android `beforeinput` capture for Backspace/Enter, the IME/Gboard commit handling, and the fail-stop that discards the queue, disarms and blurs rather than letting continued typing become a reply with an Enter attached. That last one is the detail I'd have most expected a first pass to miss.

What changed, and why, since it's a fair amount:

**The entry point moved to a named "Type" toggle in the Controls row, beside Keys.** The long press was tried, then a split Send button, and both were rejected in testing. The hold saves one tap, but the tap is per session rather than per keystroke — you arm once and type many keys — and what it buys is an accidental hold quietly pointing the phone keyboard at a live terminal. The split button cost a third of the primary action's width, permanently, for a mode used in bursts. Keys turned out to be the real sibling: it exists because a phone keyboard can't send Esc/Tab/arrows/chords, this exists because it can't send bare printable letters, so someone hunting for `b` finds it where they already looked.

**It now also disarms on a hidden page and on a composer lock** (gone pane, read-only device, idle pause), alongside your pane-switch and failed-batch cases. The mirror is polled and the poll stops in all of those, so the next keystroke would land in a terminal nobody is watching. Your fail-stop is why a lost connection didn't need its own trigger.

**Armed state gained a strip in the document flow** above the input — the restyled button and textarea are exactly what a user stops looking at once typing starts.

**Renamed from "Immediate mode"** — it named the latency rather than the destination; the UI says "Type into terminal" against "Send reply".

The 0.26.0 bump came out of the branch; main cut one release for the batch.

Still untested where you said it was: Android/Gboard. jsdom only simulates that event stream, so if you run this on a real device, swipe-typing a word and Backspace mid-word are the two paths I'd want to hear about.

aspiers:
@AltanS Thanks! Sorry for the late reply, but this works pretty well for me, including swipe-typing a word and Backspace mid-word.

The only issue I have noticed is that when I am on a tab running an agent and the "Agent" option is available in the controls at the bottom of the screen, it pushes the settings icon to the right of that almost completely off the right hand side of the screen, making it very hard to spot and tap.

PR #78, opened 2026-08-09
fix(web): make the reply guard's input-box detection width-independent

### LINKED ISSUE #76: Reply guard stalls when it can't find the input box — narrow panes and long wrapped drafts
## What happens

Sending from the PWA fails with "Message didn't reach the input box", but the text itself does land in the box — only Enter is withheld (the audit log shows the `reply` line with no `submit` line). Pressing Enter by hand over SSH delivers the message as-is.

There are two independent causes, but the path is the same: both make `locateInputBox()` return `null`.

## Cause 1: the border test requires a 20-glyph run

`isBoxBorder` (`web/src/lib/harness/claude/markers.ts:31`) requires 20+ consecutive rule glyphs. Reasonable as false-positive protection, but it silently assumes the pane is at least 20 columns wide.

Herdr's grid follows the width of attached clients, so with a narrow client attached (a 70-col SSH session in my case) split panes came out at 19 columns — a 19-glyph border, one short of the threshold. The top border can also carry an embedded session name (after resume, for example — it seems case-by-case), which splits the run, so on a narrow pane the flanking runs never reach 20 either.

## Cause 2: `MAX_DRAFT_LINES = 12`

The scan in `locateInputBox` walks up past wrapped draft lines with a 12-line cap (`web/src/lib/harness/claude/chrome.ts:32`). A 610-char / 25-line message wrapped to about 40 rows on a 41-col pane (CJK glyphs take 2 cells), so the scan gave up before reaching `❯` and returned `null`. This one happens at any width if the message is long enough, and the paste-placeholder fallback doesn't engage either, since it also requires a non-null draft.

## Proposal

Both constants guard against the same thing — stripping content from a borderless buffer — but I think that job is already done structurally by the shape `locateInputBox` requires (bottom border → `❯` → top border). So:

1. judge a border by the line's shape (a line consisting only of rule runs plus an optional short label) instead of an absolute run length;
2. scan up to the top border instead of a fixed cap.

If this direction sounds right, I'd be happy to put up a PR with tests.

**Env:** collie 0.24.2; confirmed both constants are unchanged on current `main`.

tyamanak:
Fixes #76.

## The bug

The #34 guard locates Claude's input box by the shape `bottom border → ❯ prompt → top border` (`locateInputBox`, chrome.ts). Two of its premises fail on real screens, and both failures end the same way: the box is "not found", the typed text is verifiably sitting in it, and the submit key is withheld forever.

**1. `isBoxBorder` required 20+ consecutive rule glyphs.** That's a width assumption in disguise: Herdr's grid follows the narrowest attached client, so with a 70-col SSH client attached my split panes rendered 19-glyph borders — one short. A session label spliced into the top border (`───── my session name ──`, observed flanks 5/2) splits the run, so neither flank reaches 20 even on a normal pane.

**2. The wrapped-draft scan was capped at 12 lines.** CJK glyphs occupy two cells, so a 610-char / 25-line Japanese message wraps to ~40 rows on a narrow pane. The scan gave up before reaching `❯`, and the paste-placeholder fallback couldn't help either (it also needs a non-null draft).

## Why not just lower the number / raise the cap

Any absolute run-length threshold keeps the hidden width assumption — there is always a pane narrow enough to break it. And the 20-glyph test was doing double duty as false-positive protection, so simply lowering it would let prose separators (`— — —`), table dividers (`│ │ │`), and CJK dash idioms (`―― 中略 ――`) read as borders — during review we constructed a full permission-dialog buffer that completes the border→❯→border shape out of exactly those lines, which would point the submit key at a live dialog. The protection has to come from somewhere other than a big number.

## The fix

Three layers, each with one job:

- **Glyph + shape, not length** (`isBoxBorder`): a border is the one glyph Claude actually draws (U+2500) — bare at ≥ 8 display cells, or `──…── label ──…──` with ≥ 2-glyph flanks and a label containing at least one real (non-rule, non-space) character. Em-dash prose, corner-glyph dialog borders (`╭──╮`), and rule-only "labels" all fail on glyph or shape, at any width.
- **A looser test for the top border only** (`isInputBoxTopBorder`): Claude's own renderer clamps a label's placement to `Math.max(1, …)` (traced from the bundled binary), so a real top border can carry a 1-glyph flank (`──── fast mode ─`). Step (e) is the last anchor — the bottom border, the prompt line, and the walk budget are already established — so it alone affords the looser floor. The renderer's overflow shape (flank 0) is documented as deliberately unhandled: it's lexically indistinguishable from prose, and the raw-mirror fallback is safe.
- **A 100-line walk budget, blanks included** (was 12): defense-in-depth rather than a correctness bound — it covers the observed ~40 rows with margin while keeping a stray rule line from pairing with a quoted `❯` far across the read window. Blank padding draws from the same budget (a free blank-skip was a bypass).

Floors are measured in display cells, not UTF-16 units: a CJK label halves its char count (`─ 中文 ─` is 8 cells but `.length` 6) and combining marks inflate it (`── é ──` decomposed is `.length` 8 but 7 cells). The width helper (`text-width.ts`, new, self-contained) feature-detects `Intl.Segmenter` with a code-point fallback, per 46a85d1 — module-scope construction never throws on older engines.

Every bypass shape found during review is pinned as a must-not test (the fake dialog above, the 5-col `─ x ─` mini-border, its combining-mark variant, far-alignment past the budget), alongside positives for the 19-glyph bare border, the observed 5/2-flank labelled border, 1-glyph flanks, and a 640-char contiguous-Japanese draft wrapped to 40 rows and verified through `draftCarriesSend`'s fold-seam path.

### DIFF
--- a/web/src/lib/harness/claude/chrome.test.ts
+++ b/web/src/lib/harness/claude/chrome.test.ts
@@ -4,7 +4,8 @@ import { describe, expect, it } from "vitest";
 
 import { parseAnsi } from "../../ansi";
 import { splitLines, type StyledLine } from "../../blocks";
-import { extractInputDraft, extractStatusLines, stripChrome } from "./chrome";
+import { draftCarriesSend } from "../../reply-action";
+import { extractInputDraft, extractStatusLines, hasInputBox, stripChrome } from "./chrome";
 import { lineText } from "./markers";
 
 /** The statusline run as plain text. extractStatusLines returns STYLED lines — a statusline tells
@@ -17,8 +18,9 @@ const statusText = (lines: StyledLine[]): string[] =>
 const PANES_DIR = join(import.meta.dirname, "..", "..", "..", "fixtures", "panes");
 
 // Synthesise the input-box shape: a top rule, the "❯ …" prompt line, a bottom rule, and an optional
-// statusline below it (matched by position, like the real captures). 40 box glyphs clear the
-// 20-glyph border threshold in isBoxBorder.
+// statusline below it (matched by position, like the real captures). 40 glyphs is comfortably above
+// isBoxBorder's BARE_BORDER_MIN floor (8) — see the narrow (19- and 8-glyph) and labelled-border
+// cases below for the width-dependent bug this shape used to have.
 function boxBuffer(promptLine: string, status?: string): StyledLine[] {
   const rule = "─".repeat(40);
   const rows = [rule, promptLine, rule];
@@ -31,6 +33,15 @@ function boxBuffer(promptLine: string, status?: string): StyledLine[] {
 function wrappedBoxBuffer(promptLine: string, continuationLines: string[], above?: string[]): StyledLine[] {
   const rule = "─".repeat(40);
   const rows = [...(above ?? []), rule, promptLine, ...continuationLines, rule];
+  return splitLines(parseAnsi(rows.join("\n")));
+}
+
+// Same shape as boxBuffer, but with a caller-chosen border WIDTH — for pinning bug #76's
+// width-dependent border test at a narrow (sub-20-glyph) pane.
+function narrowBoxBuffer(promptLine: string, ruleWidth: number, status?: string): StyledLine[] {
+  const rule = "─".repeat(ruleWidth);
+  const rows = [rule, promptLine, rule];
+  if (status !== undefined) rows.push(status);
   return splitLines(parseAnsi(rows.join("\n")));
 }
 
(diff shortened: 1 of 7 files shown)


AltanS:
Merged as de88b38 — thank you, and thanks for the report in #76 that came with it.

The width-independence argument is the right one: any absolute run-length threshold just relocates the assumption. Two things beyond the fix itself were worth the review on their own — closing the two unbounded `while (isBlank) i--` walks in `locateInputBox` so blank padding draws from the same budget as real continuation lines, and `text-width.ts` following `reply-action.ts`'s `Intl.Segmenter` feature-detect rather than white-screening an older engine at module evaluation.

Heads-up on one downstream edit: #79 (clipping long rule rows in the mirror) landed right after yours in #83, and its clip threshold had been justified by citing the 20-glyph `RULE_RUN` you removed. Re-anchored it to reasons of its own, and `rule-glyphs.ts` now records why the input-box grammar and the presentation rule must not be unified — your `markers.ts:86-103` note about the last time that was tried made the case.

tyamanak:
Thanks for the review and the merge!

PR #79, opened 2026-08-09
feat(mirror): clip long terminal rule rows

en-ver:
## Summary

Prevent long horizontal terminal rules from wrapping into multiple visual rows on narrow mirror viewports. In wrap-on mode, independently classified rule rows are visually clipped at the mirror edge while their complete contents remain in the DOM.

## Motivation

Terminal-width separator rows can wrap repeatedly on smaller browser viewports, adding distracting visual height and making the surrounding output harder to read.

## Implementation and scope

- Classify a row only when its trimmed content contains at least 20 repetitions of one supported horizontal Unicode rule glyph.
- Preserve outer whitespace and the complete segmented text.
- Exclude ASCII rules, mixed-glyph rows, labels, prose, corners, junctions, vertical borders, and table edges.
- In wrap-on output, constrain only classified raw rows to the mirror width and hide their visual overflow.
- Preserve ANSI styling, links, find-match spans, selection, indentation, and full DOM text without adding an ellipsis.
- Leave ordinary output on its existing wrapping path.
- Leave wrap-off output column-faithful and horizontally pannable.

This is a generic browser-presentation change. It does not change PTY or backend behavior, terminal width, dependencies, input-box detection, or agent adapters.

## Release / compatibility

Includes synchronized `0.25.1` changelog and version metadata.

## Validation

Validated on candidate `3ed0ea7`:

- Root typecheck passed.
- Root tests passed: 534 tests, including control-script lifecycle coverage.
- Web typecheck passed, including the worker configuration.
- Web tests passed: 1,618 tests across 101 files.
- Focused coverage verifies classification boundaries, DOM text and ANSI-style preservation, links and find offsets, ordinary wrapping, and wrap-off panning.
- `scripts/check-version.sh` passed.
- `git diff --check` passed.
- Production build passed with build identity `0.25.1+3ed0ea7.1786218409`.
- Production-CSS validation passed across Chromium, Firefox, and Playwright WebKit at 240px and 2000px widths with 9px, 11px, and 12px fonts.

The browser matrix verified one-row clipping, containment, preserved DOM text and selection, ANSI styling, normal prose and link wrapping, the 19/20 classification threshold, and wrap-off horizontal panning. Playwright WebKit is not native Safari; native Safari and full end-to-end PWA rendering were not tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to Collie are recorded here. The format follows
 `version` in `herdr-plugin.toml`, `package.json`, and `web/package.json` (enforced by
 `scripts/check-version.sh`). See [`CLAUDE.md`](./CLAUDE.md) → *Versioning* for the bump policy.
 
+## [0.25.1] - 2026-08-08
+
+### Changed
+
+- **Long terminal rule borders clip at the mirror edge** instead of wrapping into several rows; ordinary output keeps normal wrapping (ab23eb1)
+
 ## [0.25.0] - 2026-08-07
 
 ### Added
--- a/herdr-plugin.toml
+++ b/herdr-plugin.toml
@@ -1,6 +1,6 @@
 id = "herdr.collie"
 name = "Collie"
-version = "0.25.0"
+version = "0.25.1"
 min_herdr_version = "0.7.0"
 description = "Mobile web UI to monitor and reply to your agent herd, served over Tailscale"
 platforms = ["linux", "macos"]
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "collie",
-  "version": "0.25.0",
+  "version": "0.25.1",
   "private": true,
   "license": "MIT",
   "description": "Collie — a mobile web UI to monitor and reply to your Herdr agent herd over Tailscale",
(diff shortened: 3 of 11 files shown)


AltanS:
Landed in #83 (merged as 4480019) — thank you, this was a clean, well-scoped change and the tests came along with it.

I took it over rather than asking you to rebase, because it collided with #78 (opened two minutes after yours, so neither of you could have seen the other). #78 fixes issue #76 by **deleting** the 20-glyph `RULE_RUN` — that run turned out to be a hidden "the pane is at least twenty columns wide" assumption, and a 19-column pane stalled the reply guard on it. So your rewrite of that constant became dead code; your `RULE_ONLY` extraction and `rule-glyphs.ts` stand as you wrote them.

Your commit is cherry-picked with authorship preserved. On top of it: re-anchored the clip threshold's justification (it cited the run #78 removes, so it had gone from stale to false — twenty is still right, just for reasons of its own), wrote down at the shared surface why the input-box grammar and the clip rule must keep separate predicates and thresholds, and added a test pinning the labelled border across both modules. The 0.25.1 bump came out — main has several unreleased features queued for one release commit.

Also: #73's premise held up. Clipping the rule rows was the narrower fix.

PR #82, opened 2026-08-09
fix(web): bundle Nerd Font symbols so PUA glyphs stop rendering as tofu

### LINKED ISSUE #70: Bug: Nerd Font glyphs render as blocks on mobile
# Bug: Nerd Font glyphs render as blocks on mobile

## Bug Description
Terminal pane output containing Nerd Font private-use-area (PUA) glyphs renders as empty "tofu" boxes on mobile browsers (Android/iOS). These icons are essential for tool/agent status representations (e.g., used by OMP/herdr `symbolPreset: nerd`).

## Root Cause
- **Component:** The output is rendered via `web/src/components/ansi-output.tsx`, which uses Tailwind's `font-mono` utility.
- **CSS Resolution:** `font-mono` resolves through a CSS custom property in `web/src/index.css`:
  `--font-mono: "JetBrains Mono", "Cascadia Mono", "Cascadia Code", "Roboto Mono", ui-monospace, SFMono-Regular, "SF Mono", Menlo, Consolas, "Liberation Mono", monospace;`
- **Missing Asset:** There is no `@font-face` declaration or bundled webfont to support these glyphs.
- **Mobile Restriction:** Mobile browsers do not implement the Local Font Access API. Unlike desktop environments, they cannot resolve Nerd Font glyphs from system-installed fonts.
- **Evidence:** The repo's own test fixture `web/src/fixtures/panes/claude--trust-prompt.txt` contains PUA glyphs (e.g., U+F4B5, U+F0219, U+F02A4, U+F12E9, U+F1362) which trigger this behavior.
- **Nuance:** This bug is mobile-guaranteed; it is desktop-conditional because desktop browsers (Chrome/Edge) implement the Local Font Access API, masking the issue if the user happens to have a Nerd Font installed locally.
- **Infra Readiness:** The project's existing build infrastructure is already prepared for a fix:
  - `web/vite.config.ts` (Vite PWA `injectManifest.globPatterns`) already includes `woff2` in the precache list.
  - The bridge (`bridge/server.ts`) is configured to serve `.woff2` files correctly.
  - The CSP (`default-src 'self'`) permits same-origin font loading.
  A proper fix is a low-risk, small-footprint addition.

## Environment
- **Platform:** Mobile (Android: Chrome/Firefox; iOS: Safari)
- **Deployment:** Standard web browser and PWA install

## Steps to Reproduce
1. Open the Collie web interface on a mobile device.
2. Navigate to a pane that produces output containing Nerd Font symbols (e.g., OMP/herdr tool status icons).
3. Observe that the icons render as empty squares ("tofu").

## Expected vs Actual
- **Expected:** Icons render correctly.
- **Actual:** Icons render as empty tofu boxes.

## AI Suggested and Temporarily Working Fix
*Note: This is a local stopgap solution used to verify the issue and is not a permanent, optimized upstream fix.* This does work for me until I update collie and have to patch again.

1. **Add Font File:** Downloaded `SymbolsNerdFontMono-Regular.ttf` and placed it under `web/public/fonts/`.
2. **CSS Update:** Added the following to `web/src/index.css`:
   ```css
   @font-face {
     font-family: "NerdFontSymbols";
     src: url("/fonts/SymbolsNerdFontMono-Regular.ttf") format("truetype");
     font-display: swap;
   }
   ```
3. **Update Font Stack:** Updated `--font-mono` to prioritize the new symbols font:
   ```css
   --font-mono: "NerdFontSymbols", "JetBrains Mono", "Cascadia Mono", "Cascadia Code", "Roboto Mono", ui-monospace,
     SFMono-Regular, "SF Mono", Menlo, Consolas, "Liberation Mono", monospace;
   ```

**Caveats:**
- The `SymbolsNerdFontMono-Regular.ttf` file is ~2.2MB. A production-ready fix would ideally use a significantly smaller, subsetted webfont containing only the required PUA ranges.
- This is a local patch applied to the plugin; it is not durable against plugin updates or reinstalls.

AltanS:
Closes #70.

The font stack named no webfont, so resolving Nerd Font glyphs depended on one being installed **on the client**. Desktop browsers can do that; mobile browsers have no Local Font Access — so on the device Collie exists for, every Powerline separator, devicon and Material icon was a tofu box.

**What ships**

Two subsets of Symbols Nerd Font Mono v3.5.0 (MIT), split at the plane boundary: **626 KB** BMP private-use (`U+E000-F8FF`) + **492 KB** plane-15 Material Design (`U+F0000-F1AFF` — 6,896 of the font's 10,519 glyphs, and untouched by a Powerline-only prompt).

**What keeps it cheap**

- `unicode-range` per face: a browser fetches a file only when a codepoint in that range is actually *painted*. Measured in a real browser against the production bundle — a plain-text page fetches **zero** font files; the fixture's glyphs (U+F4B5, U+F0219, U+F02A4, U+F12E9, U+F1362) fetch both and render.
- **Out of the SW precache** (`globPatterns` drops `woff2`) — precaching would charge every install ~1.1 MB up front, including installs that never need a glyph. Cached first-use instead by a small cache-first route, no new workbox dependency.
- That cache is permanent, so it stores only an **unredirected 200 whose content type says font**: a fronting proxy answering an expired session with a sign-in page returns 200 too, and an unguarded `response.ok` would serve that HTML as the font forever — the shape of the 401 that once froze an installed SW, one layer down. The write is best-effort and off the response path; a superseded version is swept on `activate`.
- Version in the filename, since `public/` assets are unhashed and that cache never revises an entry.

**Scope of the subset:** private-use codepoints only. The characters the font also carries outside them (♥, ⚡, ☰) already render from the system font; overriding them would change text that reads fine today.

**Known and accepted:** a symbol takes this font's advance width, so an icon nudges the mirror's columns to its right by a fraction of a cell — fixing that needs a terminal emulator, which Collie does not run (ADR 0008). A device that installs the PWA and goes offline without ever painting a glyph sees tofu until it is online once. Pre-v3 configs emitting the old `U+F500-FD46` Material block still tofu — those glyphs moved upstream; widening the range can't bring them back.

`scripts/build-nerd-font.sh` regenerates the pair. It is **not** wired into the build: needing Python + fonttools to build Collie is a worse trade than 1.1 MB in git. `fonts.test.ts` pins that the stylesheet, the SW and the files on disk agree, and that `woff2` stays out of the precache.

No version bump here — the release commit that ships this will carry it.

Thanks @Shadorain for the report; the diagnosis was already most of the way to the fix, including the infra readiness notes.


Shadorain:
Sweet this is awesome! Good stuff, thank you !!

PR #85, opened 2026-08-10
fix(state): stop the session-name poll from scrolling idle claude panes

OowhitecatoO:
Every idle Claude pane scrolls up and snaps back to the bottom, once per poll, for as long as Collie runs. It is frequent, it is baffling until you know what causes it, and it moves the text under your eyes while you are reading it — none of which this read needs in order to do its job. It takes a pane shorter than 40 rows: `enrichSessionNames` reads each claude pane with `source: "recent"`, `lines: 40`, and Herdr's agent-automation docs say that for an idle, recognised agent a `recent` read asking for more rows than the visible screen "automatically use[s] the agent's mouse-scroll interface", collecting overlapping pages and "returning the viewport to the bottom before completing the read". Claude runs on the alternate screen, which has no host scrollback, so driving the agent's own scroll is the only way to reach those rows — and you watch it happen. My panes were about 70 rows on a 4K screen and this never fired once; moving to 1920x1080 took them to 26–28 rows and it fired on every poll.

The read now asks for `visible` — the rendered viewport, clamped to it however large `lines` is, so there is nothing above it to harvest. Everything else is unchanged: same line count, same extraction, same consumers of `sessionName`. The test fake now records the arguments it receives and pins the source, because this symptom is invisible to CI — it only appears on a physical terminal.

<details><summary>Details, if useful</summary>

Live-probed against herdr 0.8.0, on an idle claude pane with `viewport_rows: 71`:

| `source` | `lines` | `format` | elapsed | lines returned |
|---|---|---|---|---|
| `recent` | 70 | `text` | 0.00s | 71 |
| `recent` | 71 | `text` | 0.00s | 72 |
| `recent` | 72 | `text` | 0.85s | 73 |
| `recent` | 400 | `text` | 13.8s | 401 |
| `recent` | 200 | `ansi` | 0.00s | 71 |
| `recent` | 600 | `ansi` | 0.00s | 71 |
| `visible` | 600 | `ansi` | 0.00s | 71 |

One row over `viewport_rows` is enough. The 401-line result is the giveaway: the pane shows 71 rows and the alternate screen has no scrollback, so the other 330 came from scrolling the agent.

Confirmed blind: with Collie stopped, I fired six `visible` reads and then two `recent`/40 reads back-to-back at a pane I was watching, without knowing the counts beforehand. I saw exactly two scrolls.

`bridge/server.ts` is left alone. Its pane-mirror read is 600 lines, far past any pane height, but it passes `format: "ansi"` and no `ansi` read was ever observed to harvest; it also genuinely wants scrollback, and only runs while someone has that pane open. I left a comment there recording the dependency, since nothing else makes it visible.

The 40 stays 40. With `visible` the count no longer decides anything about scrolling, and the run below the ❯ prompt is a statusline of unknown height (ADR 0004), so the headroom is worth keeping.

`HERDR_API.md` gains the probe table and the rule it implies: a background poll must read `visible`, and anything asking for scrollback is asking to move the operator's screen.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to Collie are recorded here. The format follows
 `version` in `herdr-plugin.toml`, `package.json`, and `web/package.json` (enforced by
 `scripts/check-version.sh`). See [`CLAUDE.md`](./CLAUDE.md) → *Versioning* for the bump policy.
 
+## [0.26.1] - 2026-08-10
+
+### Fixed
+
+- **Idle Claude panes no longer scroll up and snap back on every poll** — the session-name sniffer read `recent`, which on a pane shorter than the read makes Herdr scroll a full-screen agent to reach the rows above it; it reads the visible grid now (b00e8d1)
+
 ## [0.26.0] - 2026-08-10
 
 ### Added
--- a/HERDR_API.md
+++ b/HERDR_API.md
@@ -40,6 +40,31 @@ the socket assumptions behind the design in [`ARCHITECTURE.md`](./ARCHITECTURE.m
     visible grid, and its renderer hard-wraps prose at the pane width — there are no soft-wrapped
     rows for the unwrap to merge. It only differs on scrollback-accumulating panes (shells: one
     probe measured 199 → 188 lines with logical lines up to 222 cols re-joined).
+  - **A `recent` text read can scroll the pane it reads.** Herdr's agent-automation docs state that
+    for an idle, recognized agent at the bottom of its transcript, `recent` / `recent_unwrapped`
+    reads "automatically use the agent's mouse-scroll interface" when `lines` asks for more than the
+    visible screen, collecting overlapping pages and "returning the viewport to the bottom before
+    completing the read". Claude runs on the alt screen, which has no host scrollback, so that is the
+    only way those rows can be reached — and the operator watches their terminal scroll up and snap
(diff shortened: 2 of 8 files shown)


AltanS:
Landed on `main` as dab122e — your commit, your authorship, cherry-picked with `-x` so it points back at b00e8d1. Thank you, this was a really good catch and an even better writeup.

The blind confirmation is what sold it: firing six `visible` reads and two `recent` ones at a pane you were watching, without knowing the counts, and seeing exactly two scrolls. I reproduced it here on a 49-row sandbox claude pane — `recent`/40 is 1 ms, `recent`/60 is **1212 ms**, and `visible`/600 clamps to the viewport at 102 ms. Exactly the threshold you documented. I also checked the sniffer against live named panes: `visible` extracts every `/rename` name correctly, and on a pane taller than the read `visible` and `recent` return byte-identical text, so nothing regresses.

The `bridge/server.ts` comment is the part I'm most glad you added — that the 600-line mirror read is safe *only* because it's `ansi` was a live dependency with nothing anywhere making it visible. Refusing to promise `ansi` never harvests was also right; I confirmed it in my probes but it stays an observation, not a contract.

**The one thing I dropped was your release commit** (`chore(release): 0.26.1`). Not a criticism — CLAUDE.md as written told you to bump, and you followed it exactly. But #86 landed the same day, so one 0.26.1 will cover both, and two PRs each guessing the next version collide. I've since written the actual rule down (8f4276c): **a PR from a fork leaves all four version files alone** and the maintainer cuts the release commit. That was our gap, not yours.

You'll get a mention in the 0.26.1 notes when I tag it.

PR #90, opened 2026-08-11
fix(push): Apple refuses the update topic — "collie-update" is an impossible base64 length

ojulean:
## What's broken

Every *update-available* push is rejected by APNs with
`400 {"reason":"BadWebPushTopic"}`, for every Apple subscription. Agent alerts
on `collie-herd` deliver normally, so push looks healthy — the only symptom is
that Apple operators never hear about a new release.

I hit this on a self-hosted deployment with Apple-only subscriptions, after
0.25.0 started logging the push service's real reason. Before that it was an
unexplained retry.

## Root cause

RFC 8030 §5.4 describes the collapse topic as URL-safe base64. **Apple enforces
that by decoding it**; FCM and Mozilla treat it as an opaque string. Base64 turns
3 bytes into 4 characters, so a trailing group of one character is impossible —
no valid base64 string has a length ≡ 1 (mod 4).

`collie-update` is 13 characters. `collie-herd` is 11, which happens to be fine,
which is why only half the feature broke.

Measured against a live Apple subscription, one variable at a time:

| topic | len | len mod 4 | TTL | urgency | APNs |
|---|---|---|---|---|---|
| `collie-herd` | 11 | 3 | 21 600 | high | ✅ 201 |
| `collie-herd` | 11 | 3 | **259 200** | **omitted** | ✅ 201 |
| `collie-updat` | 12 | 0 | 21 600 | high | ✅ 201 |
| `collie-updates` | 14 | 2 | 21 600 | high | ✅ 201 |
| `collie-update` | 13 | **1** | 21 600 | high | ❌ 400 `BadWebPushTopic` |
| `collie-herdab` | 13 | **1** | 21 600 | high | ❌ 400 `BadWebPushTopic` |
| `collie-update-xyz` | 17 | **1** | 21 600 | high | ❌ 400 `BadWebPushTopic` |

Those rows exist to kill the alternative explanations, since the two option sets
differ in three ways and only one matters:

- **Not the TTL** — the 3-day value succeeds on a working topic (row 2).
- **Not the missing `Urgency`** — omitting it succeeds on a working topic (row 2).
- **Not the wording** — `collie-herdab` fails identically to `collie-update`.
- **Not a length cap** — 14 characters succeeds where 13 fails.

## The fix

```ts
const UPDATE_SEND_OPTIONS = { TTL: 259_200, topic: "collie-updates" } as const;
```

One character. `collie-herd` needs no change. Nothing else in the codebase reads
the topic — the client and service worker collapse on the separate `tag`, and
`update.ts`'s `collie-update-check` is an unrelated User-Agent.

I've extended the comment above the constants, because this is exactly the
constraint someone tidies away: shortening a topic by one character can break
Apple delivery while every other service keeps working, and nothing surfaces it
but that service's log line.

## The guard

`topicIsSendable` is pure and exported for the test, following
`isReservedAuthPath` in `server.ts`. The test asserts the topics the code
**actually emits** through a capturing sender, not the constants restated — a
guard that restated them would still pass after someone edited one, which is the
only regression it exists to catch. I verified that by reverting the constant
locally: the guard fails, and passes again when restored.

The predicate keeps RFC 8030's alphabet and 32-char ceiling and adds the
mod-4 rule, so it encodes "sendable to every service", not "sendable to Apple".

## Testing

Full `ci.yml` sequence, in order, on Node 22 (the lineage `ubuntu-latest` ships)
with Bun 1.3.14:

| step | result |
|---|---|
| `scripts/check-version.sh` | ✓ 0.27.0 consistent |
| `bun run typecheck` (root) | clean |
| `bun run test` (bridge) | **548 tests across 26 files, 0 fail** — includes the `collie-ctl` lifecycle suite |
| `cd web && bun run typecheck` | clean (app + worker tsconfigs) |
| `cd web && bun run test` | **104 files, 1718 tests, 0 fail** |

The change is confined to `bridge/`; the web suite is included above for
completeness rather than because it exercises anything new here.

No version bump or CHANGELOG entry, matching how fork PRs land here — those move
in your own `chore(release):` commits. Happy to add one if you'd prefer.

---

*claude-opus-5 on behalf of Ovidiu Julean*

### DIFF
--- a/bridge/push.test.ts
+++ b/bridge/push.test.ts
@@ -3,7 +3,7 @@ import { mkdtemp, readFile, rm, stat } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
-import { Push } from "./push.ts";
+import { Push, topicIsSendable } from "./push.ts";
 import type { PushSender, PushSubscription } from "./push.ts";
 import { loadConfig } from "./config.ts";
 
@@ -287,7 +287,7 @@ describe("Push — per-message collapse topic (update must not share the herd sl
 
     // No `urgency` — an update notice is NOT worth punching through Android's power-saving the way a
     // waiting agent is. The absence is the decision; `toEqual` pins it.
-    expect(sends[0]!.options).toEqual({ topic: "collie-update", TTL: 259_200 });
+    expect(sends[0]!.options).toEqual({ topic: "collie-updates", TTL: 259_200 });
     expect(JSON.parse(sends[0]!.payload).data.target).toBe("settings");
   });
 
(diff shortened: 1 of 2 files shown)


AltanS:
Merged — the measured matrix (13 fails, 12 and 14 pass, wording-neutral) is exactly the evidence this needed. Thanks!

PR #93, opened 2026-08-11
feat(harness): give omp an adapter, so a reply stops confirming its pickers

qaz74107410:
# feat(harness): give omp an adapter, so a reply stops confirming its pickers

Adds a harness adapter for the Herdr agent string `omp` (oh-my-pi v17.2.12), which currently has no
adapter and falls through to the raw terminal mirror.

**Tier 1** — no interactive block kind is emitted, so no tap can become a keystroke:

```ts
export function ompBuildBlocks(lines: StyledLine[]): Block[] {
  return [{ kind: "raw", lines: stripChrome(lines) }];
}
```

One unconditional return. `grep -rn "kind:" harness/omp/` returns exactly that line, so a non-raw
block is unreachable by construction rather than by test luck.

## What it changes for an omp user

`reply-action.ts` opens with `if (!adapter) return oneShot(args)`, so an omp pane takes the legacy
one-shot send — type **and** submit in one call. A reply sent from a phone while one of omp's pickers
holds the keyboard fires the submit key at that picker, confirming whatever row is highlighted.
Registering an adapter swaps that for type-then-verify; this one adds `composerReady`, a pre-flight
that reads the pane before typing and refuses on a definite `false`. All eleven modal captures in the
corpus answer `false`.

Also: omp's own composer box stops drawing underneath collie's on the mirror, a laptop-typed draft
becomes readable, and omp gets a slash palette (`commandsFor("omp")` currently returns `[]`, so the
button never renders).

## One change outside `harness/omp/`

Flagging this because it touches core rather than the adapter. `composer.tsx`'s destructive pre-clear
sweep (`ctrl+k` + a Backspace run) fires *before* `sendGuardedReply` is called — before anything has
read the live pane. It decides from `display`, which lags a poll while following and is frozen while
the user has scrolled back, so a dialog opening in that gap takes the burst. This predates the branch
and isn't omp-specific; registering an adapter is just what makes it reachable on omp panes.

The change: the sweep runs only where a live read positively saw the composer, and the burst is bound
to the screen that authorised it. It's structural — the pre-flight returns a runner built on the one
branch where `composerReady` answered true, so `force`, a read that threw, and an adapter without
`composerReady` all get `null` and skip it. The binding reuses the mechanism `dialog-guard.ts` already
uses (`expected_prompt`, with the bridge's 409 `prompt_changed`). Full reasoning is in
`fix(reply): a destructive pre-clear needs a live read, and a binding to it`.

**Happy to pull this out into its own PR** if you'd rather review it separately, or drop it entirely
and ship just the adapter.

## Verified on a live pane (2026-08-11)

Real omp v17.2.12 pane, Herdr 0.8.0, bridge built from this branch, driven through the real
`sendGuardedReply` + `lib/api.ts` + `adapterFor("omp")`, with `composer.tsx`'s `send()` mirrored
line-for-line as the caller. Every request logged, with the bridge's `audit.log` as an independent
record.

| Case | Result |
|---|---|
| Empty composer | No sweep, no second read; typed, verified, then submit. omp answered. |
| Stranded draft | Sweep bound to the `╰─ … ─╯` row; guard re-read at +382 ms; submitted text verbatim, no concatenation. |
| `/model` picker up | Refused `blocked`. One request total — the pre-flight GET. Pane byte-identical before/after (`cmp`: 0 differing bytes over 26 179). |
| `force` retry, picker up | Typed without sweeping (no `/keys` request at all); submit key withheld; no row confirmed. |
| Binding race, staged | Bridge refused 409 `prompt_changed`; zero `/reply` requests; the burst never reached the pane. |

Not covered: one pane, one omp version, one terminal width, one modal shape. The tool-approval dialog
wasn't exercised (see below). I didn't reproduce the old `oneShot` behaviour on `main` for comparison,
since that would have confirmed a `/model` row and changed a persistent setting.

## Corpus

Twenty byte-faithful captures across three sandbox panes, never a real session: 9 composer states, 11
with a modal owning the screen. No picker's confirm key was pressed — every dialog was driven onto the
screen, captured, and dismissed with `Escape`.

Sanitized in place, length-preserving, so column alignment and display width survive byte-for-byte and
the captures stay usable as width fixtures. Rewritten: the environment (cwd, MCP server names, session
titles, a machine-local `skill:` row) and vendor account state (which providers the account is signed
into). Preserved: the shape the detectors read. The corpus README publishes the greps that verify it.

I went with sanitizing since the captures came off a real machine into a public repo, and there's
precedent in the existing corpus. Let me know if you'd rather have them a different way and I'll
recapture.

## What this deliberately doesn't include

- **omp's tool-approval dialog** — the Tier-2 candidate, and why I split this. It's also **not in the
  corpus**, so to be precise about what's proven: `raw` is returned unconditionally, so it can't be
  up-levelled either way — but the claim that `composerReady` answers `false` on it is *inferred* from
  the eleven modals that were captured, not measured. Capturing it is the first thing a Tier-2
  follow-up owes.
- **A generic `menu` lift** for omp's pickers — `parseKeyHintFooter` returns `[]` for nine of omp's
  footers (omp writes `<key> <verb>`, not `<key> to <verb>`), and widening the shared grammar would
  move a contract the existing `/model` picker is pinned against. Left as a tripwire test instead.
- **The `ask` tool's select dialogs** — declined on the fail-closed contract.
- **A thorough conformance run.** With `ownFixtures: []` the suite emits nine `it.todo`s. What
  executes is conservative detection (raw-only over all 58 captures, in full) plus the new
  `composerPrompt`/`composerReady` pairing. The real gating for this adapter is `omp/chrome.test.ts`,
  and the code says so rather than letting a green run imply more than it proves.
- **A fully closed pre-clear window.** The bridge's re-read and its `send_keys` are two separate herdr
  RPCs, so a dialog opening between them is unguarded — the same residual the other guarded keystroke
  paths carry. Bounded: the submit key is still withheld, so the worst outcome is an unsubmitted
  message.
- **A proof that `╰─ … ─╯` is the composer's alone** — it's a census (once in each of the 9 composer
  captures, nowhere in the other 49), not a proof, and worth re-checking against a future omp version.

## Tests

Against a pristine `main` export, same runner and `node_modules`: zero new failures, +364 passing
(`48 failed | 1670 passed` → `48 failed | 2034 passed | 10 todo`). The failing-file list is identical
and the 48 failing test *names* were diffed rather than counted — byte-identical sets, all the
pre-existing `localStorage is not available` environment fault on my machine. `tsc --noEmit` clean in
both packages; root `bun run test` 545 pass / 0 fail. Capability fence green. No `.skip`, `.only`, or
`TODO` placeholder in any new or modified file.

One edit to `HARNESS_CONTRIBUTING.md`: the CI-gate bullet, since the conformance suite gained a leg.
`CHANGELOG.md`, `package.json` and `herdr-plugin.toml` are untouched — the version seemed yours to
pick.

First contribution here, so tell me if any of this is shaped wrong and I'll rework it.


AltanS:
Excellent work, especially the reply-path fix: the live-read gate closes a hole that predates omp, and the review confirmed it's safer for every harness. I pushed two small commits onto your branch before merging: d76a4c1 pins down a test that could leak its guard poll into later tests, and 2c44255 has composerPrompt decline a binding the bridge's 6-row tail window could never verify (a tall slash palette would have made every send 409). Thanks!

qaz74107410:
Thanks for the merge, and for the two follow-ups. The `composerPrompt` one especially: the corpus had exactly two palette captures, at 3 and 5 rows below the box, and I read the 5 as headroom when it was the ceiling. `suggestEnd - bottom - 1 > 5` is the check I should have derived from `verifyExpectedPrompt` rather than from what the captures happened to show.

Issue #95, opened 2026-08-12
Plan feedback has no route from the phone — and every answer button is a silent no-op while the dialog's input is focused


navidkashani:
On the plan-approval dialog, row `4. Tell Claude what to change` is hidden by Collie and the composer
refuses to send while the dialog is up — so **plan feedback has no route from the phone**.

While probing that I found something worse, which is most of this report: **the dialog has states in
which every button Collie renders is a silent no-op**, and one in which it renders a button labelled
with whatever the desktop user is halfway through typing.

Everything below was measured on a scratch pane one keystroke at a time, not read off a screenshot.
Environment: Collie 0.28.0 (`2910f40`), Claude Code **2.1.228**, herdr, iOS PWA over Tailscale.

---

## 1. Row 4 is not an option — it is a text input. Hiding it is correct.

Worth stating first, because it means `isFreeTextLabel` is doing the right thing and the fix is not
"stop hiding row 4".

`4. Tell Claude what to change` is an **inline text input**, and that string is its *placeholder*.
With the box empty, pressing `4` answers nothing — it moves `❯` onto the row and focuses the field:

```
 ❯ 1. Yes, clear context (4% used) and use auto mode   │    1. Yes, clear context (4% used) and use auto mode
   2. Yes, and use auto mode                           │    2. Yes, and use auto mode
   3. Yes, manually approve edits                      │    3. Yes, manually approve edits
   4. Tell Claude what to change                       │  ❯ 4. Tell Claude what to change
      shift+tab to approve with this feedback          │       shift+tab to approve with this feedback
              (baseline)                               │            (after pressing `4`)
```

So a phone tap on a "4" button could never have answered this dialog.

## 2. The gap: two correct rules, closed at the same moment

- `isFreeTextLabel` drops the row because it is answered by **typing**, and the composer covers typing.
- The composer's pre-flight refuses to send because a dialog is on screen and the input box isn't.

Each is right alone. Together they mean the one dialog whose free-text row is the entire point is the
one where typing is unavailable.

## 3. The dialog has four states, and Collie only renders the first one correctly

The pointer position and the box contents vary independently, and **both** change what Collie
produces and what a tap does:

| # | State | What Collie renders | What tapping a button does |
|---|---|---|---|
| 1 | pointer on 1–3, box empty | 3 correct buttons | answers correctly ✅ |
[...]

Issue #89, opened 2026-08-10
Unable to change mode with Shift + Tab


dantebarba:
Hello!

Sending the Shift + Tab command does not change modes in Claude Code, at least using the Keys pad. I'm connecting via Safari to the PWA.

dantebarba:
Quick Diagnosis by Claude 

> Bottom line: herdr's pane.send_keys is broken for shift+tab — it reports success and delivers nothing the TUI acts on. Collie is just relaying that call, which is why the Keys pad can't work no matter what you tap. Your physical keypress works because the terminal encodes the key itself.
> 
> What works today:
> 
> herdr pane send-text w7:p1 $'\x1b[Z'    # cycles one step: manual → accept edits → plan
> 
> From the phone specifically, there's no path through Collie's Keys pad until one of these lands:
> 
> 1. herdr fixes send_keys — clean repro is send-keys shift+tab (no effect) vs send-text $'\x1b[Z' (works). Worth filing; it's a one-line encoding bug.
> 2. Collie adds a "cycle mode" button that calls the text endpoint with \x1b[Z instead of send_keys. That's sendKeys in web/src/lib/api.ts plus a preset in the Keys pad — you've forked a plugin before, and this is smaller than that one.

AltanS:
Confirmed and root caused. Herdr delivers shift+tab as a bare Tab (0x09) instead of ESC [ Z, so Claude never sees the back-tab. Collie relays the key faithfully, so nothing to fix on our side. Already fixed upstream (herdrdev/herdr#1561, PR herdrdev/herdr#2259), merged the day after 0.8.0 shipped, so it lands in the next herdr release. Until then `herdr pane send-text <pane> $'\x1b[Z'` cycles the mode. Keeping this open until that release is out.

AltanS:
Fixed upstream in Herdr 0.8.2 (herdrdev/herdr#1561, PR #2259). Verified here with a live sandbox probe: shift+tab now correctly cycles Claude's mode indicator (manual → accept edits → plan → auto), matching what the physical keypress does. Thanks for the root cause and the workaround while we waited.

PR #96, opened 2026-08-12
fix(harness): the plan dialog's feedback row is an input, and focus changes what digits mean

navidkashani:
Fixes the detection half of #95.

`isFreeTextLabel` matches the plan-approval feedback row's **placeholder**, so it only recognises
that row while the box is empty. The row is an inline text input, not an option — pressing `4` moves
`❯` onto it and focuses it, answering nothing — so dropping it is right. But the placeholder holds
only while it is empty:

| Box state | Row 4 label | `isFreeTextLabel` |
|---|---|---|
| empty | `Tell Claude what to change` | ✔ matches → dropped |
| typed | `use a guard clause instead` | ✘ **misses** → up-levelled into `keys:["4"]` |

So once anyone types into it, Collie renders a live button labelled with that person's half-written
sentence. The state survives moving the pointer away (`4`, type, `Up`), so it isn't transient.

**Fix 1 — key the row off its description.** `shift+tab to approve with this feedback` comes from a
static `description` field and is present in both states. This needs each row's description collected
*before* the free-text check rather than after it.

**Fix 2 — decline the dialog while that input has focus.** This is the one I'd flag for review, since
it removes buttons that render today. While `❯` sits on the input row, Claude Code routes **every
digit into the field as text** instead of answering. Measured on 2.1.228:

```
before                                       after send_keys ["3"]
   3. Yes, manually approve edits               3. Yes, manually approve edits
 ❯ 4. Tell Claude what to change              ❯ 4. 3
      shift+tab to approve with this feedback       shift+tab to approve with this feedback
```

The plan is not approved and nothing surfaces on the phone. In that state the three answer rows still
parse as an ordinary menu, so lifting them puts buttons on a phone that are indistinguishable from
working ones and that silently type into a sentence someone else is mid-way through writing. Bailing
to the raw mirror + keys pad is the same call the `>9 options` ceiling already makes: a keystroke that
cannot fire is worse than no button.

### Evidence

Two fixtures, captured live on Claude Code **2.1.228** in a scratch pane driven one keystroke at a
time, banner truncated so nothing identifying is in them:

- `claude--plan-approval--feedback-focused.txt` — pointer on the input, box empty. Detection must
  return `null`.
- `claude--plan-approval--feedback-typed.txt` — text typed, pointer arrowed back to row 3. Upstream
  returns four options here, the fourth `{ label: "use a guard clause instead", keys: ["4"] }`; with
  this change it returns the correct three.

The choreography behind them is written up as `web/src/lib/grammar/PLAN_FEEDBACK_NOTES.md` in the
format of `NOTES_NOTES.md` — the full key table, what `Enter` and `Esc` actually deliver (read out of
the agent's transcript, not inferred from the screen), and three hazards for anyone automating this
later. Notably the footer is **not** usable as a focus discriminator here: `ctrl+g to edit in Vim` is
present in every state, the opposite of the notes flow.

Both fixtures are registered in the places the repo's guards require — `chrome.test.ts`'s `PINNED`
table and `DIALOG_FIXTURES`, `prompt-binding-regions.json`, and the panes README table. The focused
one is in `conformance.test.ts`'s `NEUTRAL` cohort, with a comment noting it is the one entry there
that *is* a dialog and is declined deliberately.

### Verification

- `bunx vitest run` (web) — **2164 pass**, 10 todo, 107 files
- `bun test ./bridge ./scripts` — **570 pass**, 0 fail
- `bun run typecheck` — clean, root and web
- `scripts/check-version.sh` — `✓ version 0.28.0 consistent` (no version or CHANGELOG files touched,
  per the fork-PR rule in CLAUDE.md)

Not live-verified on a phone — I don't want to claim a Tier-2 confirmation that's yours to make. The
grammar change is fixture-driven and the fixtures are real captures.

### Scope

This is detection only. It deliberately does not add a way to *send* plan feedback from the phone:
the verified sequence is `4` → text → `Enter`, which gives **deny-with-feedback** (the agent keeps
planning). Approve-*with*-feedback needs `shift+tab`, which is #89. Where such an affordance should
live, and how the block should behave while the sequence is mid-flight, is a UI decision I'd rather
leave to you than guess at — I've put the verified choreography in the notes file so it's there if
you want it.

### CHANGELOG line

```
### Fixed
- Plan-approval feedback row: identified by its description, so a typed-in box no longer becomes a live button carrying the user's own text; the dialog is declined outright while that input has focus, where every digit is swallowed as text (f200c7b)
```


AltanS:
Merged via #102 — your commit is in as-is (`c0ce09e`), authorship intact. Thank you for the report and the fixtures; the keystroke-by-keystroke measurements were what made the rest of it possible.

Two things came out of building on it: the row is now modelled rather than dropped when focused (so the phone can say *why* the buttons are dead, and the feedback flow's polls can observe the focus they wait for), and — the one your fixtures couldn't have caught — the row **re-flows** a long value across several lines instead of windowing it, which broke the description marker and pushed the footer past `MAX_FOOTER_GAP` hard enough to drop the whole dialog to the raw mirror at 355 characters. All measured and written up in `PLAN_FEEDBACK_NOTES.md`, which is still substantially yours.

PR #99, opened 2026-08-15
 feat: add AGY (Antigravity) harness adapter, interactive prompts, and slash commands

Kryvonis:
## Summary

Adds support for **AGY (Google Antigravity CLI / `agy` / `antigravity`)** to Collie:

1. **Interactive Prompt Up-Leveling (`web/src/lib/harness/agy/`):**
   - Lifts `ask_question` dialogs, numbered options (`1. Option`, `1) Option`, `[1] Option`, `( ) 1.Option`), and permission/confirmation prompts into native tappable buttons.
   - Handles options with or without explicit footer hint lines.
   - Guards against free-text sends while an interactive prompt is open (`composerReady`).
   - Re-derives signatures through the race guard before sending keystrokes (`[digit, "Enter"]` or `[digit]`).

2. **Slash Command Palette (`web/src/lib/agent-commands.ts`):**
   - Adds the public AGY command catalog with prefix matching for `agy`, `agy-cli`, and `antigravity`:
     `/plan`, `/grill-me`, `/learn`, `/compact`, `/clear`, `/model`, `/status`, `/subagents`, `/skills`,  `/rules`, `/mcp`, `/hooks`, `/review`, `/diff`, `/doctor`, `/theme`, `/help`, `/exit`.

3. **Brand Styling & Icon (`web/src/components/agent-icon*`):**
   - Adds the official brand mark and tile background (`#1A73E8`) for `agy` and `antigravity` pane badges and switcher tiles.

4. **Registry & Case/Prefix Normalization (`web/src/lib/harness/registry.ts`):**
   - Normalizes agent keys in `adapterFor` to support casing and common prefix variants (`agy`, `agy-cli`, `antigravity`).

    ---

## Suggested CHANGELOG Entry

AGY (Antigravity CLI) gets a harness adapter — up-levels interactive question options (`ask_question`) and tool permission dialogs into tappable buttons, gates the composer against typing into active menus, adds the AGY slash command palette, and includes brand icons.

 ## Verification

    - **Harness Unit Tests:**
      `cd web && bunx vitest run src/lib/harness/agy/` (7 passed)
    - **Commands & Icon Tests:**
      `cd web && bunx vitest run src/lib/agent-commands.test.ts src/components/agent-icon.test.tsx` (39 passed)
    - **Full Frontend Suite:**
      `cd web && bun run test` (108 test files, 2,161 tests passing)
    - **Bridge Tests:**
      `bun test bridge/` (559 tests passing)
    - **Typecheck & Production Build:**
      `cd web && bun run typecheck && bun run build` (0 errors, clean build)
    - **Live Local Test:**
      Verified prompt button tapping and submission against a live `agy` session on Herdr.


AltanS:
Thanks — the agy work is welcome. The command catalog and the icon are fine as they are.

The adapter emits `prompt-select`, whose keystroke recipe is already live, so this is Tier 2 in [HARNESS_CONTRIBUTING.md](https://github.com/AltanS/collie/blob/main/HARNESS_CONTRIBUTING.md) and needs: a captured fixture corpus, a notes file with the verified recipe, a `describeAdapterConformance` call, and live verification. I can't do the last one — agy isn't installed here.

Two changes I'd want regardless:

1. `keys: [n, "Enter"]` synthesises digits from numbered rows, and the detector matches any numbered list near the tail — it'll claim `/model` and `/skills`. [ADR 0009](https://github.com/AltanS/collie/blob/main/.adr/0009-a-generic-menu-is-driven-by-the-keys-it-names.md) closes that off (in Claude's `/model` a digit also rewrites the user's default). Drive it from the footer, or record the live probe showing agy's digits are safe.

2. Please drop the `adapterFor` prefix-matching — it widens claude/omp to any agent string with those prefixes. Separate PR, separate argument. Register agy by its exact agent strings.

Bring the corpus + notes + conformance and run the live check on your session, and I'll take it whole.

AltanS:
Both asks are addressed — `adapterFor` is exact-keyed again, and the detector is footer-gated so `/model` and `/skills` stay raw. Thanks. Three things before I take it.

**1. The fixtures aren't captures.** Every file in the corpus is a byte-faithful `format:ansi` capture with real ESC bytes; the 8 agy files have none, and their rule widths disagree between files (60 vs 113 columns).

```
claude--select-menu.txt  27 ESC     grok--ask-color.txt   54 ESC
codex--ask-fruit.txt     78 ESC     agy--select-menu.txt   0 ESC
```

This is the one that blocks the rest: the keystroke table in `AGY_NOTES.md` is the only evidence the recipe is right, and hand-typed screens can't corroborate it. Capture them with `scripts/capture-fixture.sh <paneId> <name>` off a sandbox agy pane. If you need to scrub a username or path, keep the substitution length-preserving so the row padding stays byte-identical — see the codex corpus note in the fixtures README.

**2. `chrome.ts` never finds the input box.** `? for shortcuts` matches its `PROMPT_REGEX` (the `\?\s*` branch), so the walk stops on the status line. On your own fixtures:

```
agy--fresh-idle.txt  DRAFT = "for shortcuts    accept-edits · Gemini 3.7 Flash · low"
agy--done.txt        DRAFT = "for shortcuts    accept-edits · Gemini 3.7 Flash · low"
agy--fresh-idle.txt  blocks = [{raw, 9 lines}]      // the box is still in the mirror
```

Collie would paint that as the operator's draft. `extractStatusLines` returns `[]` unconditionally too, so the status row is never lifted out. Please cover both against a fixture — the unit tests use hand-built strings, which is why this got through.

**3. Please restore the completeness gate in `prompt-binding-contract.test.ts`.** Narrowing it to `claude--` also switches it off for codex, grok and omp. With the original filter, five agy fixtures come back unpinned:

```
agy--permission-bash, agy--permission-edit, agy--plan-approval,
agy--select-menu, agy--trust-prompt
```

Claude's `detectPromptSelect` claims all five. Add their regions to `prompt-binding-regions.json` rather than shrinking the filter.

Two notes, not conditions. `isAlienBuffer` is carrying more weight than it looks: with that one function stubbed out, the agy grammar claims 12 foreign fixtures with digit keys. Nothing routes that way in production, so it isn't a live hazard — but it tells you the grammar is loose, and a real agy screen that prints `Enter to select` will be claimed the same way. A brand-name blocklist won't catch that one. Separately, the `gemini` → `someunknown` edit in `agent-icon.test.tsx` looks unrelated; `gemini` is still an unknown agent everywhere else.

Bring the real captures and the chrome fix and I'll take it.

Kryvonis:
Thank you @AltanS
First of all for this powerful plugin and for maintaining it all. 
I'm glad to commit and help a bit
alt

---
Addressed your comments so please review whenever you have free minute, and feel free to reach out to me

PR #100, opened 2026-08-15
Name a pane by its terminal title, so a project's herd stops reading as N identical rows

praneetrohida:
## The problem

Run several Claude agents in one space and the herd list gives you no way to tell them apart. The title line is `project · tab`, Herdr names an unlabelled tab positionally, and the second line falls back to a cwd that is identical on every row. Three agents in one project render as three copies of the same row.

The two names Collie already had — `pane.rename` and Claude's `/rename` — are both set once, by hand. Neither tracks what the agent is actually doing.

## What this does

Reads the pane's OSC title, which Herdr has been reporting all along.

It arrives **on the pane object**, not only on `session.snapshot`'s `agents[]`, so this needs no change to the "agents stay derived from `panes`, one code path" decision — `agents[]` stays unused exactly as before. `WirePane` simply never declared the field, so it was dropped at the type boundary on every poll.

Precedence becomes `paneLabel` → `sessionName` → `terminalTitle` → cwd. The hand-set names still win: a name you chose shouldn't be overwritten by one the process rewrites every turn.

## Why not just use `terminal_title_stripped`

Because it doesn't finish the job, and the half it misses is the animated half.

Herdr strips the settled `✳` but leaves Claude's rotating spinner frames. Measured over 450 polls of a live 3-agent herd on herdr 0.8.0: `✳ <title>` was stripped every time, `◐ <title>` and `◑ <title>` never were. Those frames advance on each poll, so binding a label straight to that field makes every working row's name flicker.

So the strip is Collie's own. It whitelists the symbol blocks a spinner is drawn from (Geometric Shapes, Misc Symbols, Dingbats, Misc Symbols and Arrows, emoji, plus bare `*`/`·`) and requires the trailing space that always follows a status glyph. Requiring that space is what lets it handle frames nobody has observed yet without eating a title that merely *opens* with punctuation — `(main) vim`, `— draft`. Herdr's stripped form is still used as a head start when it's the shorter of the two, so an older server that omits it takes the same path.

A title is dropped when it only repeats the row: the agent name (what Herdr falls back to when nothing sets a title), or the project already on line one. Same shape and same reasoning as `meaningfulTabLabel`, and it lives beside it.

## Verification

- `bun test ./bridge ./scripts` — 578 pass
- `web`: `vitest run` — 107 files, 2145 pass
- `bunx tsc --noEmit` clean in both trees
- Built and run against a live 3-agent herd: each row renders its own task name, no glyph, stable across polls

## Notes

- No version bump and no CHANGELOG entry, per `CLAUDE.md` § Versioning for fork PRs.
- Suggested CHANGELOG line, if you want one:
  > **A pane is named by what its process says it is doing** — Herdr's OSC title fills the second line where the cwd used to repeat the project, so several agents in one space stop rendering as identical rows. Herdr's own `terminal_title_stripped` leaves Claude's rotating spinner frames in place, so the glyph strip is Collie's.

### DIFF
--- a/bridge/activity.test.ts
+++ b/bridge/activity.test.ts
@@ -7,6 +7,7 @@ import {
   ActivityLedger,
   coerceActivityFile,
   meaningfulTabLabel,
+  meaningfulTerminalTitle,
   PRUNE_AFTER_MS,
 } from "./activity.ts";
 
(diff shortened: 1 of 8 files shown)


AltanS:
Thanks — this held up well end to end (the OSC-title premise checked out on live data here). One addition landed directly on the branch (982b8e1): a shell pane's default title is bash's `\u@\h:\w` locator (`altan@bluefin:~/projects/…`), which was replacing the row's cwd line with a longer restatement of the same cwd and renaming the pane on every `cd`. `meaningfulTerminalTitle` now drops that shape as another 'repeats what the row already shows' case; command titles (`vim foo.ts`, `htop`) survive. Merging; ships in the next release.

PR #107, opened 2026-08-17
audit: optional COLLIE_AUDIT_CONTENT=none, keeping the event and dropping the body

shuangwangnyc:
audit: optional COLLIE_AUDIT_CONTENT=none, keeping the event and dropping the body

The audit trail records every write-level action, which is right — sending
keystrokes to a pane is arbitrary code execution and who-did-what-when is worth
keeping. It also records up to 120 characters of the message itself, and up to
120 characters of the prompt binding, which is a slice of whatever happened to be
on the terminal at the time.

For a personal deployment that is fine. For one where the trail is wanted but the
bodies are not, there is currently no way to have one without the other.

This adds one config value:

  COLLIE_AUDIT_CONTENT=preview   (default, byte-for-byte today's behaviour)
  COLLIE_AUDIT_CONTENT=none      every string inside `detail` becomes the length

The envelope is untouched either way: ts, action, paneId, session, device, and
every boolean and number inside detail. So submitted, textDelivered and
promptBinding.checked/passed still answer the questions the log is for.

Two small decisions, in case they are worth arguing about:

- It keeps the LENGTH rather than a shorter preview. A truncated secret is still
  a secret, and the question a reader asks of a redacted trail is usually whether
  anything was sent at all, which a count answers.
- It recurses, so nesting is covered. The prompt binding is nested and holds the
  terminal fragment; a redaction reaching only the top level would have left the
  more sensitive half in place.

sanitize and formatAuditLine take the mode as a defaulted parameter, so every
existing caller and test is unaffected.

Tested: the existing suite passes unchanged (590 tests), plus four new cases
covering the default being untouched, the envelope surviving, nothing of the
message or the terminal surviving at any depth, and the length being preserved.

Happy to adjust the naming, or drop the env var and take a constructor argument
only, if you would rather not add configuration for this.

### DIFF
--- a/bridge/audit-redaction.test.ts
+++ b/bridge/audit-redaction.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, test } from "bun:test";
+import { formatAuditLine } from "./audit";
+
+const entry = {
+  action: "reply",
+  paneId: "w1:p1",
+  session: "default",
+  detail: {
+    text: "deploy the thing and here is a secret nobody should keep on disk",
+    submit: true,
+    promptBinding: { checked: true, passed: true, expected: "user@host ~/work %" },
+  },
+};
+
+describe("audit content redaction", () => {
+  test("preview is unchanged — the default must not move", () => {
+    const line = JSON.parse(formatAuditLine(entry, 0));
+    expect(line.detail.text).toContain("deploy the thing");
+    expect(line.detail.promptBinding.expected).toContain("user@host");
+  });
+
+  test("none keeps the envelope and drops every string body", () => {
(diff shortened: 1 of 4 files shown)


AltanS:
Thanks @shuangwangnyc — this was a real gap and the shape you picked (`COLLIE_AUDIT_CONTENT=preview|none`, envelope kept, bodies dropped) is exactly what landed. Merged via #111 with your commit cherry-picked in, authorship preserved; the follow-up on top changes the redaction to decide by field role with a fail-closed allowlist (so key names stay legible while anything operator- or screen-originated redacts by default) and swaps the exact length for a flat `⟨redacted⟩`. It'll ship with the next release.

PR #108, opened 2026-08-17
fix(direct-typing): say the mode stopped when the app comes back

enieuwy:
## The bug

Backgrounding the page disarms **Type into terminal** — correctly: the mirror has stopped tracking the pane, so the next keystroke would land in a terminal nobody is looking at.

It disarmed *silently*. `clearMode()` ran with no message, and a status published while the document is hidden expires unseen anyway (`lib/status.ts` drops a non-error after 2.5s). It also left the field **focused with the keyboard up**.

So: you switch apps mid-command, come back to something that looks exactly like the mode you left, keep typing — and the characters go into the **reply draft**. A different action, with an eventual Enter attached to it.

Reproduce on a phone: arm Type into terminal, switch apps, come back, type. On `main` the text silently becomes a reply.

## The fix

- **Defer the notice to the next `visible` event.** A message published while hidden is gone before anyone can read it.
- **Mount the visibility listener for the hook's whole life**, not keyed on `active`. Hiding the document is *what disarms*, so an effect keyed on the armed state tore its own listener down in the same commit — the return trip had nobody left to report it.
- **Put the keyboard away on disarm.** The notice expires; a focused field does not. A primed field outlasting the only warning is exactly how terminal keystrokes become a buffered reply.
- **Own the deferred blur.** It has to be deferred (`focusInputEnd()` is itself a timeout, so a synchronous blur is undone by a focus already queued), which means it can land after the field has changed hands. It now carries the arm generation it was scheduled for, no-ops once a re-arm has taken over, and is cancelled at a pane boundary and on unmount. The failed-batch path had the same unowned deferred blur and now shares the guarded one.
- **Write the armed flag at the transitions**, not during render. The listener fires between renders; a ref that catches up on the next one can misreport which state it is describing. `resetMode()` is now the single teardown, shared by `clearMode`, the failed-batch callback, the background disarm and the pane boundary.

## Tests

7 new tests in `composer.test.tsx`, each failing before the change:
- the deferred announce on return,
- blur-on-background,
- the notice surviving to a visible page,
- a stale blur not reaching into a re-armed session.

Each was proved to catch the old behaviour by reverting the hook under the new test.

`bunx tsc --noEmit` clean. `composer.test.tsx`: 79 pass / 2 fail — the same 2 fail on `main` (`restores the draft after a remount`, `keeps drafts per pane`); they're localStorage-less-environment failures, not from this change.

## Versioning

Per `CLAUDE.md`, fork PRs leave the version files and CHANGELOG alone. Suggested entry if you want one:

> - **"Type into terminal" now says when it stopped.** Backgrounding the app disarms the mode; it did so silently and left the field focused, so the next thing you typed became a reply draft instead (#NN, `452da20`)


AltanS:
Merged — thank you. The diagnosis was the valuable part: a status published while the page is hidden expires unseen, so the disarm had to announce itself on the way back rather than on the way out, and blurring the field matters as much as the message does.

Two follow-ups landed on top (1a2ca49). The arm-generation token was removed — `activate()` already cancels the pending blur and is the only re-arm path, so the token could not differ by the time the timer ran, and the test written for it passed against the unfixed hook too. Ownership is settled by cancellation alone now, and the test fails without it. And `backgrounded` is now cleared at the pane boundary: a pane can change while the page is hidden (push notifications deep-link into one), which had the return trip announcing the disarm over a pane the mode was never armed on.

Ships in the next release; no bump needed on your side, and you had the fork rules exactly right.

PR #115, opened 2026-08-18
feat(composer): add speech input through Codex-owned auth

ardaaltinors:
## Summary

Adds optional speech input to the composer for users who already have an authenticated Codex installation.

A microphone button appears immediately to the left of Send. Recorded speech can either:

- be sent through Collie’s existing guarded reply path; or
- be inserted into the composer as an editable draft.

No OpenAI API key, separate OAuth flow, provider configuration, model selector, or OpenAI SDK is added to Collie.

## Why revisit speech input after #91?

#91 was declined because it made Collie responsible for transcription credentials, provider configuration, an outbound path, and an additional SDK dependency.

This implementation deliberately narrows that ownership boundary:

- Codex remains the owner of authentication and token refresh.
- Collie communicates with the locally installed `codex app-server`.
- Collie never reads or writes Codex auth files or the system keychain.
- Refresh tokens are never exposed to Collie.
- Access tokens exist only in bridge process memory.
- No new runtime dependency is introduced.
- If a usable Codex ChatGPT login is unavailable, the feature remains visible but disabled.

Collie still owns microphone capture and an outbound audio request when the user explicitly records. That remaining tradeoff is documented rather than hidden.

## User experience

- Adds a microphone control immediately before Send.
- Tap to start recording; tap again to transcribe.
- Recording can be cancelled without submitting audio.
- Maximum recording duration is 2 minutes.
- Maximum recording size is 25 MB.
- Recording is cancelled if the user changes pane/session or leaves the composer.
- Preferences are stored per device, following Collie’s existing browser preference pattern.

Two settings are provided:

### Speech to text

Automatically available when an authenticated Codex ChatGPT session is detected. Users can disable it. Without compatible authentication, the setting is greyed out and cannot be enabled.

### Hands free

Enabled by default.

- Enabled: the transcript is sent through the existing guarded reply flow.
- Disabled: the transcript is inserted at the current caret position and is not sent.

Existing typed text is not replaced by a hands-free recording. If submission fails, the transcript is retained as a draft.

## Architecture

The recorder and UI depend on a small internal `SttProvider` interface. This PR ships only a Codex implementation; it does not add other providers or generic provider configuration.

The Codex provider:

- maintains one long-running `codex app-server` process;
- obtains short-lived authentication through its compatibility RPC;
- accepts only ChatGPT authentication, not API-key login;
- asks Codex to refresh after a 401 and retries exactly once;
- supports both current nested and older flat ChatGPT account ID claims;
- serializes transcription requests;
- rejects redirects;
- applies bounded timeouts and sanitized error reporting.

The transcription route uses Collie’s write-access guard. Provider availability uses the read-access guard.

## Deliberately out of scope

- Independent OAuth or token storage
- Reading or modifying `~/.codex/auth.json`
- OpenAI API-key configuration
- Generic endpoint/model settings
- Additional cloud or local STT providers
- Local inference
- Continuing a recording after its originating composer is unmounted

The provider boundary leaves room for future implementations without making them part of this change.

## Privacy and stability

Audio is sent from the bridge host to ChatGPT only after the user starts and completes a recording.

The Codex transcription endpoint is private and undocumented, and the app-server interface is experimental. Either may change upstream. The feature fails closed and becomes unavailable when the Codex integration cannot be used.

## Verification

Automated:

- Bridge test suite: 653 passing
- Codex STT compatibility suite after the account-claim fix: 8 passing
- Web test suite: 2,346 passing, 10 existing todos
- Focused composer suite: 5 passing
- Root and web TypeScript checks passing
- Production build passing
- Version consistency check passing

Manual:

- Tested with an existing Codex ChatGPT login on macOS
- Provider reported available through the running Collie bridge
- A real 2.2-second OGG recording was accepted and transcribed successfully
- Verified both hands-free submission and draft insertion behavior

## Fork versioning

Per the contribution guidance, this PR does not modify `package.json` versions or `CHANGELOG.md`.

Suggested changelog entry:

> **Codex speech input.** When an existing Codex ChatGPT login is available, the composer can record a reply and either send it through the guarded reply path or keep it as an editable draft, without Collie owning OAuth credentials.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,9 @@ COLLIE_HOST=127.0.0.1
 # Usually injected by the systemd unit; override only if your socket is elsewhere.
 # Defaults to ~/.config/herdr/herdr.sock, or %APPDATA%\herdr\herdr.sock on Windows.
 # HERDR_SOCKET_PATH=/home/you/.config/herdr/herdr.sock
+
+# Optional absolute path when Codex is installed outside the service's PATH (common with nvm/launchd).
+# COLLIE_CODEX_BIN=/home/you/.local/bin/codex
 # Which dialer opens that socket: auto (default) | net | bun. "auto" is right everywhere — node:net
 # on Windows, where herdr's socket is a named pipe, and Bun's native transport elsewhere. Forcing
 # "net" on Linux/macOS runs the Windows dial path against your real socket; that's how it gets
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@
 A phone web UI for your [Herdr](https://herdr.dev) agent herd, served over Tailscale. Open a URL, see
 which agent is waiting on you, and answer it with your phone's keyboard.
 
-The reply box is an ordinary text field, so your phone's own voice dictation works in it; Collie
-ships none of its own.
+The reply box is an ordinary text field, and can optionally record a voice reply through an existing
+Codex ChatGPT sign-in.
 
 It assumes a [Tailscale](https://tailscale.com) tailnet — your phone and the host on the same one —
 and it is **single-user**: one operator, one tailnet, no multi-tenant auth. If you need shared or
@@ -24,9 +24,10 @@ public access, Collie isn't built for it. Read the
 - **Special-keys pad** — `Esc`, `Ctrl+C`, arrows, combinable modifiers
 - **Find in output**, and **conversation history** the terminal can't scroll back to
 - **Send an image** from your camera roll
+- **Speech to text through Codex** — record a reply without copying OAuth credentials into Collie
 - **Switch between Herdr sessions** without touching the host
 - **Installs to your home screen** (PWA) and runs entirely on your own machine — loopback bind, no
-  cloud, no account
+  Collie account (the optional Codex transcription sends audio to ChatGPT)
 
 ## Contents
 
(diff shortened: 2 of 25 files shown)


AltanS:
Thanks for this, and for taking #91's objection seriously rather than resubmitting it. The
narrowing is real work: no new dependency, no key handling, no OAuth of your own, the write guard on
transcription and the read guard on availability, and the serialised refresh so two clients cannot
race. I still have to decline it, and the reason is not the size.

`bridge/stt/codex.ts` posts to `https://chatgpt.com/backend-api/transcribe` with the operator's
ChatGPT token and these headers:

```ts
"User-Agent": "codex_cli_rs/0.0.0 (Collie)",
originator: "codex_cli_rs",
```

That tells OpenAI's backend the request is the Codex CLI when it is not. The PR is upfront that the
endpoint is private and undocumented, but not about who carries the risk: we would ship the
impersonation and the user's ChatGPT account would absorb any rate limit, flag or ban. I am not
willing to put your users there, however carefully the rest is built.

Separately, the bridge still gains an outbound network path, now carrying microphone audio, plus its
first long-running child process spawned from `PATH`. The outbound path was most of why #91 was
declined, and that part has not changed.

One thing worth knowing if you build this elsewhere: there is no `isSecureContext` or
`mediaDevices` check in the diff. The button is gated on bridge-side provider availability, so on a
plain-HTTP Collie it would appear enabled and then fail to record.

The request itself clearly makes sense, and yours is the second PR for it, so I have opened
https://github.com/AltanS/collie/discussions/118 to work out what people actually need and what
shape could land. Your answer on hands-free sending would be genuinely useful there, since #91 went
the other way and kept Send explicit.

PR #120, opened 2026-08-19
feat(nav-tray): F1–F12 behind an F keys disclosure

### LINKED ISSUE #119: Nav tray: no way to send function keys (F1–F12) — harness keybinds are unreachable from the phone
### The gap

The mobile Keys tray (`nav-tray.tsx`) covers Esc/Tab/arrows/Enter/Space, sticky modifiers, Ctrl presets and a digit pad — but no function keys. Herdr's `pane.send_keys` grammar accepts `F1`…`F12` bare (verified, per HERDR_API.md), and harnesses genuinely bind them to actions. Concrete case: the pi coding agent's CE Workflow extension registers

- F7 → open Orchestrator
- F8 → Microcompact
- F9 → Fleet view

Those work from a physical keyboard but are simply unreachable from the phone, because the tray has no F-key buttons to stage or fire them.

### Suggested fix

An "F keys" disclosure in the Keys tab, mirroring the existing Presets collapsible: collapsed by default (zero tray growth), expanding to a 4×3 grid of F1–F12 rendered through the same `navBtn` path — so the press echo, the key-queue compose mode, and chord composition with armed modifiers (`ctrl+F7`, …) all come for free.

Long-term alternative/complement: user-defined key macros (label + keys array, per install), so environment-specific hotkeys like the CE Workflow trio can become first-class labeled buttons instead of being hardcoded upstream.

### Workaround in the meantime

I patched my install's `nav-tray.tsx` with a 3-column F7/F8/F9 row and confirmed it works end-to-end through the normal `pane.send_keys` path — Collie → Herdr → pane → the pi extension's F7 handler opens the Orchestrator, exactly like the physical key. Happy to open a PR for the collapsible F1–F12 version.

martin-tahli:
Closes #119.

## The gap

The mobile Keys tray had no function keys. Herdr's `pane.send_keys` grammar accepts `F1`…`F12` bare (HERDR_API.md), and harnesses bind them to real actions — e.g. the pi agent's CE Workflow extension registers F7 → Orchestrator, F8 → Microcompact, F9 → Fleet. From a physical keyboard those work; from the phone they were unreachable.

## The fix

An **F keys** disclosure in the Keys tab, mirroring the existing **Presets** collapsible:

- **Collapsed by default** — the tray's resting height is unchanged; you only pay for F keys when you ask for them.
- Expands to a **4×3 grid of F1–F12**, rendered through the ordinary `navBtn` path, so everything the tray already does comes for free:
  - press echo (fill + ✓ on accept),
  - key-queue staging when composing,
  - chords with armed modifiers (`ctrl+F7`, `alt+F3`, …) — all twelve keys also become reachable *as modifiers' bases*, not just bare.

## Tests

Two cases added to `nav-tray.test.tsx`, both green (33/33 in the file, `tsc --noEmit` clean):

- F keys stay behind the disclosure; expanded, F7/F12 fire as bare `["F7"]` / `["F12"]`.
- An armed Ctrl + F7 tap stages `ctrl+F7` for review instead of firing — the compose path applies to F keys like any base key.

## Notes

- Verified end-to-end on a live install first (a hand-patched F7/F8/F9 row): Collie → `pane.send_keys ["F7"]` → Herdr → pane → the extension's F7 handler. This PR is the generic version of that patch.
- The issue also floats a longer-term idea — user-defined key macros per install — which this deliberately doesn't attempt; the disclosure is the minimal generic fix.

### DIFF
--- a/web/src/components/nav-tray.test.tsx
+++ b/web/src/components/nav-tray.test.tsx
@@ -368,6 +368,36 @@ describe("NavTray", () => {
     expect(onSend).toHaveBeenCalledExactlyOnceWith(["ctrl+d"]);
   });
 
+  // ── Function keys (#119): F1–F12 behind their own disclosure, same fire/stage path as base keys ──
+
+  it("F keys stay behind their disclosure; expanded, F7/F12 fire as bare keys", async () => {
+    const user = userEvent.setup();
+    const onSend = vi.fn();
+    render(<NavTray onSend={onSend} />);
+
+    // Collapsed by default — the tray's height is unchanged until you ask for F keys.
+    expect(screen.queryByRole("button", { name: "F7" })).toBeNull();
+    await user.click(screen.getByRole("button", { name: "F keys" }));
+
+    await user.click(screen.getByRole("button", { name: "F7" }));
+    await user.click(screen.getByRole("button", { name: "F12" }));
+    expect(onSend.mock.calls).toEqual([[["F7"]], [["F12"]]]);
+  });
+
+  it("an armed modifier composes with an F key — Ctrl + F7 stages ctrl+F7 for review", async () => {
+    const user = userEvent.setup();
+    const onSend = vi.fn();
+    render(<NavTray onSend={onSend} />);
+
+    await user.click(screen.getByRole("button", { name: "Ctrl" })); // arm → composing
+    await user.click(screen.getByRole("button", { name: "F keys" }));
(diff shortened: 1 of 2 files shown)


AltanS:
Merged as-is — verified locally (full web suite + typecheck) since fork CI doesn't run here. Clean work: the disclosure pattern, the navBtn path, and the tests were exactly right. Ships with the next release.

PR #123, opened 2026-08-23
feat: Grok Build first-class harness adapter

kennymcavoy:
## Summary

Grok Build panes become first-class like Claude: composer chrome stripped with an opaque status strip re-surfaced natively, probed dialogs lifted into tappable buttons, a journal adapter, and the Grok slash catalog.

- **Chrome (Tier 1)** — `web/src/lib/harness/grok/`: the rounded composer box is located by its bottom status border and stripped; the status run is opaque across models/efforts (never matches `Grok 4.6` / `(high)`). `composerReady` gates the reply path; the startup screen's bare `[stable]` chip is matched literally as chrome so a fresh pane accepts its first message. Grok's full-width background paint on blank rows is dropped (no color inference) so the mirror doesn't stripe; glyph rows keep every background.
- **Permission cards** (`prompt-select`, family permission) — classified, not layout-pinned: the footer must name the family and count its rows (`1/N:select` + `Tab:next option` + `Ctrl+o:always-approve` + `Ctrl+c:cancel`), the last row must be the reject, the row above it the one-shot Yes, and every earlier row must prove itself a persistent mode change — those never become buttons. Unclassified rows refuse the whole card. Both live layouts (3-option rm, 4-option file-edit) lift; digits probed per layout (PERMISSION_NOTES.md), including the one-shot-doesn't-persist negative control.
- **ask_user_question** (family select) — radio cards lift with per-row digits (digit submits directly — probed). The `z` free-text row is modelled as feedback and locks the buttons whenever the keyboard sits on it; the inner hint's `Enter:edit` verb is the tell even when the row repaints idle (live-probed: a digit there types into the field). Esc-parked cards send `Tab` — the footer's own key — before the digit, because a bare digit is silently swallowed in scrollback (probed both ways). Checkbox cards stay raw (a digit submits rather than toggles). Multi-question wizards step through as consecutive cards.
- **Plan approval** (`menu`) — footer-named keys only, no digits (ADR 0009); composer gated while the review is up.
- **Journal** — `bridge/journal/grok.ts` reads `chat_history.jsonl` under `$GROK_HOME` (override `COLLIE_GROK_ROOT`) through `containedRealpath`, refusing non-UUID refs.
- **Neutral-model additions** (called out per your #99 note about separate arguments): `PromptFeedback.purpose` — Grok's `z` row is `"free-text"`, so the renderer offers no feedback composer and `submitPromptFeedback` refuses it (the Claude plan-feedback choreography is the wrong recipe); and `PromptOption.keyLabel` — the badge shows the digit when a recovery key precedes it. Claude behavior unchanged, pinned by the existing claude suites.

## The #99 bar

- Exact agent string only: `adapterFor` stays `Object.hasOwn` over `adapter.agent`; `grok-build` catalog folding lives in `canonicalAgent`, never in the registry.
- Dated fixture corpus (25 `grok--*.txt`), notes files with the verified recipes (PERMISSION/ASK/PLAN_NOTES.md), `describeAdapterConformance` running grok against the claude+omp cohorts and vice versa.
- Live verification: run by me against real Grok Build 1.0.5 panes per the arrangement in #99 — every dialog exercised end-to-end through the bridge (buttons pressed via the guarded `/keys` path, files created/removed on disk as the negative controls), plus stale-signature replays refused with 409.
- No digits synthesized from numbered rows: permission/ask digits are probed specific grammars; the plan menu is driven by the keys its footer names.

## Honesty notes

- The six Tier-1 composer files (`grok--fresh-idle` … `grok--user-bubble`) are sanitized structure fixtures (plain UTF-8, no ESC); the live 3-SGR-run border shape is pinned separately in `grok/markers.test.ts` from a dated probe. All dialog captures are byte-faithful `format:ansi`.
- Deliberately not lifted (each needs its own capture + probe): checkbox asks (digit submits, not toggles), typing a custom `z` answer from the phone, and Grok's cancel-turn panel.
- Observed variance: after a plan approval Grok sometimes gates the plan's own writes with a permission card and sometimes doesn't — both paths handled.

## Suggested CHANGELOG entry

Grok Build gets a first-class harness adapter — strips the composer chrome, re-surfaces the status strip, lifts permission cards, ask_user_question radios/wizards and plan approval into native buttons, and reads the Grok session journal.

## Verification

`cd web && bun run typecheck` clean; full web suite + conformance green; `bun test bridge/` green; live pass against real panes as above. Versions untouched at 0.32.0 per the fork-PR rule.

https://claude.ai/code/session_01PFH7jQVDVhPGC6FqLetNfd


AltanS:
Reviewed — same verdict as #124: strong shape, fail-closed classification, and the journal containment is actually stricter than our own adapters (re-running `containedRealpath` on the cached path is a nice touch). Version files correctly untouched. All gates pass here (root 687, web 2971, both typechecks clean).

Three changes requested before merge:

1. **Agent-string match.** `operator-scope.ts:39` folds `grok*` → `grok` for the slash catalog, but `harness/registry.ts` and `bridge/journal/registry.ts` match `"grok"` exactly. If Herdr reports the agent as `grok-build` on any install, that pane gets the slash palette and nothing else — no chrome strip, no dialogs, no journal. Can you confirm the exact `agent` string Herdr reports for your Grok Build panes? If it can be `grok-build` (or varies), widen the harness + journal registries to the same fold the catalog uses.
2. **One byte-faithful composer capture.** Six of the Tier-1 fixtures are sanitized (no ESC), as you disclosed. `markers.test.ts` pins the real 3-SGR border, which covers part of the gap, but the chrome strip is regression-guarded against partly synthetic input. One `format:ansi` capture of a real composer (idle + one with a draft) would close it.
3. **`scripts/journal-probe.ts`** — the per-root rework is good but unrelated to the adapter; please split it into its own commit inside this PR so it can carry its own changelog line.

Smaller, take or leave: the `free-text` entry in `prompt-select-block.tsx:63-73` carries five empty-string fields gated only by convention — a discriminated union would turn misuse into a type error.

On our side: Tier-2 means we live-verify the four dialog recipes against a real Grok Build pane before merge. Plan: merge #124 first, then this one after a rebase (six trivial overlaps).

kennymcavoy:
Review follow-up:

1. **Agent string.** Confirmed live 2026-08-23: Herdr's kind list is `grok`, and every Grok Build pane on this host reports `agent: "grok"` / `source: herdr:grok` (`pane.agent` and `agent_session.agent`). No `grok-build` variant showed up. Registries stay exact; the `grok-build` catalog fold in `canonicalAgent` is unused here. Noted on `harness/grok/index.ts`.

2. **Byte-faithful composer.** Replaced `grok--fresh-idle.txt` and `grok--draft-single.txt` with `format:ansi` captures from a sandbox pane (empty idle box, then a `testing stuff` draft). Darwin temp-dir token is length-preserved. The other four Tier-1 chrome files remain structure fixtures.

3. **journal-probe split.** `fix(journal): probe each root on its own` is now its own commit ahead of the adapter.

AltanS:
Merged — thanks. I did the post-#124 rebase on our side: union of the four both-add files, plus one semantic conflict git missed (main renamed the no-adapter test agent codexAgent → opencodeAgent; grokAgent now builds from that — f1a7dbc). Follow-up 7d8ffb4 makes codex and grok treat each other's corpora as foreign.

PR #124, opened 2026-08-23
feat: Codex first-class harness adapter

kennymcavoy:
## Summary

Codex CLI panes become first-class: the boxless chrome stripped and re-surfaced natively, and the three probed dialogs lifted into tappable buttons. Independent of the Grok adapter PR — cut straight from `main`, no shared new model fields.

- **Chrome (Tier 1)** — `web/src/lib/harness/codex/`: Codex paints a bare `› ` prompt row that word-wraps onto indented continuation rows, a dot-separated status row beneath, and exactly one blank row between sections. The composer is located by the status row at the tail (the ` · Context N% left|used` token behind two opaque `·`-separated fields — model names and paths are never matched) with the prompt row above; a transcript echo of a submitted message paints the same `› ` prefix and never has the status row beneath it. Wrapped drafts rejoin with single spaces (verified against the typed original). Known safe limit, documented: a custom or disabled `tui.status_line` never matches, so those configs fall back to the raw mirror with replies refused.
- **Folder-trust prompt** (family trust) — the exact two-option layout lifts; digits confirm directly (2 quit on the spot, 1 accepted through the guarded send path — both probed, TRUST_NOTES.md).
- **Exec approval** (family permission) — classified: the first row must be exactly `Yes, proceed`, the last exactly `No, and tell Codex what to do differently` (post-shortcut-strip), and every row between must prove itself persistent (`don't ask again …`) — those never become buttons; any unclassified or suffix-altered row refuses the whole card. Digits probed with negative controls: 1 ran the approved command, 3 rejected it with the command verified never run on disk (Codex still paints a `• Ran <cmd>` transcript row for the cancelled call — documented rendering quirk). Only the exec approval is captured; patch/MCP approvals have different headers and fail closed until probed. Config note: with `approvals_reviewer = "auto_review"` eligible requests route through a reviewer subagent and no dialog paints — captures used `-c approvals_reviewer=user`.
- **request_user_input** (family select) — `Question X/Y` cards lift one digit per row; a digit answers the current question, advances a multi-question set, and submits on the last (all probed, including the auto-added `None of the above` row). The notes-box state refuses to raw (`tab or esc to clear notes` footer or the `› Add notes` row): a digit there would type into the box. Esc interrupts the whole conversation and is never emitted.

## The #99 bar

- Exact agent string only (`agent: "codex"`); variant spellings pinned raw in the registry test.
- Dated corpus (10 `codex--*.txt`, Codex v0.149.0), three notes files with the probe tables, `describeAdapterConformance` running codex against the claude+omp cohorts and vice versa.
- Live verification: run by me against a real pane per the arrangement in #99 — trust accepted, a reply typed-verified-submitted through the new draft extraction, an approval landing its command on disk, and an ask answer arriving in the tool result, every press through the guarded `/keys` path with its region signature.

## Privacy

The corpus is sanitized with length-preserving substitutions (row padding stays byte-identical): username, hostname, the Darwin per-user temp-dir token, and the Codex session UUIDs from `codex resume` lines.

## Suggested CHANGELOG entry

Codex CLI gets a first-class harness adapter — strips the boxless composer chrome, re-surfaces the status row, and lifts the folder-trust prompt, exec approvals and request_user_input question cards into native buttons.

## Verification

`cd web && bun run typecheck` clean; full web suite + conformance green; `bun test bridge/` green; live pass as above. Versions untouched at 0.32.0 per the fork-PR rule.

https://claude.ai/code/session_01PFH7jQVDVhPGC6FqLetNfd

### DIFF
--- a/web/src/components/agent-chat.test.tsx
+++ b/web/src/components/agent-chat.test.tsx
@@ -364,14 +364,15 @@ describe("AgentChat — prompt-select race guard wiring (frozen {text, revision}
 // The block grammars are provably scoped to the pane's own adapter (spec T8): an agent with no
 // adapter gets the plain raw mirror — no prompt-select buttons, no chrome stripping, no re-surfaced
 // status strip — because running Claude-tuned matchers on an unverified TUI could mis-lift or
-// mis-strip its output. codex is such an agent; omp has an adapter but lifts no dialog kind at all.
+// mis-strip its output. opencode is such an agent (codex graduated to its own adapter); omp has an
+// adapter but lifts no dialog kind at all.
 describe("AgentChat — block-grammar scoping (an agent with no adapter)", () => {
-  // A codex agent sharing the Claude fixture's ids, so only the agent kind differs from the default.
-  const codexAgent = { ...fixtureAgents[0]!, agent: "codex" };
+  // An opencode agent sharing the Claude fixture's ids, so only the agent kind differs from the default.
+  const opencodeAgent = { ...fixtureAgents[0]!, agent: "opencode" };
 
-  it("does NOT lift a codex tail menu into buttons — it stays raw mirror text", () => {
-    renderChat({ text: MENU_TEXT, agent: codexAgent });
-    // No native prompt buttons: the Claude prompt-select grammar never runs for codex…
+  it("does NOT lift an adapterless agent's tail menu into buttons — it stays raw mirror text", () => {
+    renderChat({ text: MENU_TEXT, agent: opencodeAgent });
+    // No native prompt buttons: the Claude prompt-select grammar never runs without an adapter…
     expect(screen.queryByRole("button", { name: "Yes" })).not.toBeInTheDocument();
     // …and the menu row shows verbatim in the raw mirror instead (drivable by the keys pad).
     expect(screen.getByText(/1\. Yes/)).toBeInTheDocument();
@@ -392,8 +393,8 @@ describe("AgentChat — block-grammar scoping (an agent with no adapter)", () =>
     expect(screen.queryByText(/❯/)).toBeNull(); // the input box was stripped off the mirror
   });
 
-  it("leaves a codex input-box buffer fully raw — no status strip, box kept in the mirror", () => {
-    renderChat({ text: STATUS_TEXT, agent: codexAgent });
+  it("leaves an adapterless agent's input-box buffer fully raw — no status strip, box kept in the mirror", () => {
+    renderChat({ text: STATUS_TEXT, agent: opencodeAgent });
     // The statusline is NOT hoisted into an app strip — it stays inside the raw <pre> mirror…
     const status = screen.getByText(/\[Opus 4\.8\] ~\/webapp · main/);
     expect(status.closest("pre")).not.toBeNull();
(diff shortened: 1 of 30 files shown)


AltanS:
Reviewed — this is a strong contribution: fail-closed, tail-anchored detectors, exact-label matching, conformance cohorts extended in all three directions, and the version files correctly left alone. All gates pass here (root 669, web 2709, both typechecks clean).

Two changes requested before merge:

1. **`MAX_DRAFT_ROWS = 8`** (`web/src/lib/harness/codex/chrome.ts:31`) — a draft that wraps past 8 rows makes `locateComposer` return `null`, so the composer goes dark and the app reports "a dialog may be waiting", which is wrong. It fails safe, but the operator's text is stranded. Please raise the bound or anchor the `›` search without the cap.
2. **`approval.ts:79`** — `startLine` swallows the whole option run, so the persistent `2. …don't ask again…` row vanishes from both the buttons and the raw mirror; the phone shows digits 1 and 3 with no sign of 2. The mirror should keep showing the row even if it isn't offered as a button.

Two smaller notes, take or leave: `ask.ts:83` uses `family: "select"` whose documented contract is digit-then-Enter, while Codex submits on the digit alone — a one-line note at the family doc in `prompt-model.ts` that `option.keys` wins would keep the enum honest. The `tui.status_line` caveat in `markers.ts:33` deserves an operator-facing docs line eventually; fine to leave that to us.

On our side: the Tier-2 bar means we live-verify the dialog recipes against a real Codex pane before merge — that's on us, not you. Plan is to merge this PR first, then ask for a quick rebase of #123 (they overlap in six files, all trivially).

kennymcavoy:
Review follow-up:

1. **`MAX_DRAFT_ROWS`.** Raised from 8 → 100 (same bound as Grok/OMP/Claude). A wrap past 8 rows no longer makes `locateComposer` return null.

2. **Persist row.** Approval `startLine` now starts at the reject, so `2. …don't ask again…` stays in the raw mirror. Buttons are still only the one-shot Yes (`1`) and No (`3`).

AltanS:
Merged — thanks, this is exactly the #99 bar. Both follow-up fixes verified here (669 root / 2711 web, typechecks clean).

AltanS:
Maintainer-side Tier-2 live pass is done (codex-cli 0.145.0 + a scratch 0.149.1, real TUI, stubbed model): folder trust, exec approval (both directions, disk-verified negative control, persist row stays text-only in the mirror) and request_user_input all PASS; the uncaptured update-available card correctly stayed raw. Three follow-ups worth your eyes, none blocking:

1. The markers.ts KNOWN LIMIT is wider than the comment says: any codex that can't populate the `Context` status fields (failed auth handshake, not just a custom `tui.status_line`) loses the whole Tier-1 lift and the pane refuses replies. Worth widening the note — and maybe locating the composer without the Context token so a degraded codex keeps a phone composer.
2. `PLACEHOLDER = "Ask Codex to do anything"` isn't stable on 0.145.0 (it rotates hints); 0.149.1 matches. A minimum-supported-version line in the notes would settle it.
3. Trivia for APPROVAL_NOTES: the tool codex advertises is `exec_command` with a `prefix_rule` array, not `shell`.

request_user_input needed `-c features.default_mode_request_user_input=true` on these builds — also worth a line in ASK_NOTES.

PR #125, opened 2026-08-23
Prototype: operator-declared launchers in launchers.toml (one tap → throwaway Space)

enieuwy:
This is a prototype more than a proposal — I built it for my own phone, it has been running on my install for a day, and I am posting it mostly to ask whether the shape is right. Happy for it to be closed, or to be told the whole idea belongs in a fork.

### What I wanted

On the desktop I have a herdr popup bound to a key that runs a small TUI, I glance at it, and I quit. From the phone I could not have that. As far as I can tell a popup is invisible to Collie by construction — it has no pane id and is not in herdr's pane API, so `pane.list`/`pane.read` cannot see one. A pane is the only surface the phone can read, so the nearest equivalent I could think of was a pane that closes itself: herdr already drops a tab whose last pane closes and a Space whose last tab closes, so a command that closes its own pane leaves nothing behind.

### What this adds

`launchers.toml`, next to `commands.toml` and `keys.toml`:

```toml
[[launchers]]
command = "showy-quota-peek"   # required; typed verbatim
label = "Quota bars"           # optional; defaults to the command's first token
cwd = "~/dev/collie"           # optional; defaults to home, leading ~ expanded
```

I deliberately did **not** reach for a new env var — #109 → #112 settled where operator rows live, and this reuses that machinery rather than reopening it: same `operator-file.ts` reader, same mtime-checked live reload, same hold-the-last-good-rows posture, same drop-the-row-never-the-file validation.

A tap creates a Space labelled from the row, types the command, sends Enter, and navigates into the fresh pane with the existing `freshPane` flow.

`POST /api/launch` takes a command *string* and runs it only on exact equality with a configured row. My reasoning for a route rather than doing it client-side: `/api/workspace` plus a reply into the new pane would put the command line on the phone and leave nothing to match against, whereas this way the client can only name a row. Same write guard, same session scoping, same audit posture as the other structural creates — `command` is not added to `METADATA_KEYS`, so it redacts under `COLLIE_AUDIT_CONTENT=none` like any other content-bearing detail.

Enter is sent literally rather than `COLLIE_SUBMIT_KEYS`, since that is the submit sequence for an agent's composer and this is a bare shell prompt. A failed send rolls the Space back.

### Screenshots

Rows are `rumen-peek`, `showy-quota-peek`, and a `lazygit` row that exists only to show `cwd`.

| Dashboard section | Folded | The same rows as a sheet |
| --- | --- | --- |
|  |  |  |

The pane header at 390 px — Find, the launcher, the status badge:

One tap later, and then the part I actually care about: while it runs it is an ordinary Space, and after quitting it is gone.

| Launched | `SPACES (7)` while running | `SPACES (6)` after quitting |
| --- | --- | --- |
|  |  |  |

(Images live on an orphan `assets/launcher-screenshots` branch in my fork, so they are not in this diff. Say the word and I will re-upload them as attachments instead.)

### Where I am least sure

- **Two surfaces may be one too many.** The dashboard section is the one-tap path; the sheet exists because from inside a pane the section costs Home → tap → Back. I would happily drop either.
- **A separate file, or a row kind in `commands.toml`?** I chose separate because a command row addresses an existing pane and has a `scope`, while a launcher creates the pane, so `scope` would be meaningless for it — `keys.toml` felt like the precedent. Easy to fold in if you disagree.
- **Whether `/api/launch` should exist at all**, per the reasoning above. This is the decision I would most like checked.
- **No `confirm = true`** for launcher rows, though command rows have it. Trivial to add if you want the symmetry.
- **No ADR**, since I do not think this closes an option anyone argued for. If you think the "own file" call needs one, I will write it.
- I have only run this on **macOS**, against one herdr session. The screenshots and the geometry checks are headless Chromium at phone sizes; the day of real use is my own phone, so consider anything about other platforms untested.

### Checks

`bun test ./bridge ./scripts` → 666 pass. Web suite → 114 files / 2393 tests pass. Both typechecks clean. New tests cover the grammar (defaults, `~` expansion, each drop case, later-row-wins, a non-array `launchers`, a non-table row among good ones), the route (unlisted command rejected without creating anything; a listed row creates a Space with that label and cwd and types the command plus Enter; a send failure closes the created pane), and the two UI surfaces.

Per CLAUDE.md I have left the version files and `CHANGELOG.md` alone.


AltanS:
Thanks this is a cool addition!! I'm going to check the exact proposal and its shape in the upcoming days and get back at you.

enieuwy:
Happy to contribute to collie! You've built something great 👏

AltanS:
Thanks @enieuwy. Merged as d3de982. We cherry-picked your commit with
your authorship kept, then changed a few things on top:

- The launch waits for the new pane's shell before it types, and one tap
  is one launch.
- The rocket is out of the pane and Space headers. The launcher lives on
  the dashboard and in the "Switch pane" sheet.
- A launcher's cwd is optional. Absent means here: your home dir from the
  dashboard, that pane's own folder from the switcher.
- Tapped from a pane, a launch opens a tab beside it instead of a new
  Space. From the dashboard it still opens a Space.
- Rows read per host over the pack link, so on a pack they come from the
  machine whose row you tapped, never the lead's own file.
- A new /api/launchers route replaces the config field.

You can close this as merged by cherry-pick. The merge commit is d3de982.

PR #127, opened 2026-08-23
feat(windows): support full host deployment

kimjunny:
## Summary

- add Windows-native bridge/config/path handling and declare Windows in the plugin manifest
- replace duplicated launcher behavior with one TypeScript ctl entry point across Task Scheduler, systemd, and launchd
- harden start/stop/update/serve/log handling, including pinned update commits, fail-closed version checks, exact process targeting, and supervisor fallback
- add Windows CI, cross-platform ctl coverage, ADRs, and operator documentation

## Prior Windows direction

This intentionally revisits #71, where official Windows support was declined because lifecycle work would have landed twice without a Windows environment to test it. This branch removes that duplication by making `scripts/ctl/main.ts` the shared lifecycle implementation and adds a Windows CI job. It also overlaps the narrower headless Task Scheduler fix in #121; if this lands, the supported TypeScript backend replaces that contrib-only launcher path.

## Verification

- `bun run typecheck`
- `bun test ./bridge ./scripts` — 776 passed, 0 failed
- `bun test scripts/ctl` — 95 passed, 0 failed
- `cd web && bun run typecheck`
- `cd web && bun run test` — 2,382 passed, 10 todo
- `bun run build` — production PWA build, 20 precache entries
- `contrib/windows/collie-ctl.test.ps1` — passed
- live Windows Task Scheduler `start`, `status`, `url`, and `logs`
- live tailnet-only Tailscale Serve HTTPS request — HTTP 200 with CSP and build headers

The POSIX shell lifecycle permission check cannot pass on Git Bash/NTFS because NTFS reports mode `644`; the Ubuntu CI job runs `bun run test:ctl-posix` on a POSIX filesystem.

## Security and release notes

- remains loopback-only behind one managed Tailscale Serve front door; no Funnel
- preserves exact managed-mapping teardown and narrows Windows force-kill to the absolute bridge entry point
- keeps the repository at 0.32.0 with no fork-side release/version bump

Post-implementation review passed goal, QA, code-quality, security, and context/history lanes.


AltanS:
Thank you for the scale of work here. Full Windows host support is something we want, so a note on target and timing before you invest more:

1. **`main` is the wrong base now.** This PR is built around `scripts/ctl/**`, and the `v1` branch (which becomes 1.0) has replaced that layer entirely: every verb now lives in `cli/*.ts` behind a compiled `bin/collie`. Merging to `main` would ship code the 1.0 cutover deletes.
2. **The ask:** retarget to `v1` and rework the control layer against `cli/`. Your supervision and Task Scheduler design carries over; it is the invocation layer underneath that changed.
3. **Honest expectation:** 1.0 will not gate on Windows host support. This lands in a 1.x release. See `CONTRIBUTING.md` for the base-branch policy and gates.

If you would rather wait until 1.0 ships and rebase once against a stable `cli/`, that is a reasonable call too. Say so and we park it without closing it.

AltanS:
closing due to inactivity

PR #129, opened 2026-08-24
fail-closed Host and identity gates, loopback bind, parse .env, sniff uploads

bartholomewtj:
These are the extra security gates from my fork, ported onto current main. Not a merge of the fork — no Files tab, desktop mode, Traces, or Grok.

## What changes

**Host validation is on by default and fails closed.** A DNS-rebound `Host: evil.example` is rejected. `collie-ctl.sh` injects discovered Tailscale MagicDNS name + IPs as `COLLIE_TAILSCALE_HOSTS`, so a normal tailnet install needs no extra config. Behind your own reverse proxy you must set `COLLIE_PUBLIC_HOSTS`. Opt out with `COLLIE_ALLOW_ANY_HOST=1`.

**`COLLIE_TRUSTED_USER` rejects a missing header as well as a mismatch.** `tailscale serve` injects no `Tailscale-User-Login` for tagged nodes, so the old missing-header pass let any tagged node write. `COLLIE_SKIP_SERVE=1` (no injector) and `COLLIE_TRUSTED_USER_OPTIONAL=1` (host-local dev) keep the old pass.

**A non-loopback `COLLIE_HOST` refuses to start** unless `COLLIE_ALLOW_NON_LOOPBACK_BIND=1`. Non-loopback TCP peers are rejected. The identity, device and same-origin gates are all client-settable on a wide bind.

**`collie-ctl.sh` parses `.env` as key=value** instead of `source`. A `.env` with `$(…)` or backticks used to run as the operator on every verb and at login. systemd `EnvironmentFile=` already parsed; bash now matches.

**Uploads are typed by magic bytes**, not the client-supplied Content-Type. The old MIME lookup treated `__proto__` / `constructor` as a valid type.

**An unversioned managed checkout pins `update` to the newest release tag**, never `origin HEAD`. The in-major tag path from 0.32.0 is unchanged.

## Compatibility

A default `collie-ctl.sh start` on a tailnet should keep working (hosts are injected). Breaking cases:

- `COLLIE_TRUSTED_USER` set, caller has no `Tailscale-User-Login` (tagged node, or curl without the header)
- running the bridge without `collie-ctl.sh` and without `COLLIE_PUBLIC_HOSTS` / `COLLIE_TAILSCALE_HOSTS` — only loopback Host is accepted
- `COLLIE_HOST=0.0.0.0` without the new escape hatch
- a `.env` that relied on shell syntax

Version is 0.33.0 because the Host and identity defaults changed. Retag if you would rather cut this yourself.

## Tests

- `bun test ./bridge/config.test.ts ./bridge/server.test.ts ./bridge/uploads.test.ts`
- `scripts/collie-ctl.test.sh` — includes `test_env_is_parsed_not_executed` and the unversioned-checkout fallback now pinning to a tag


AltanS:
The CI failure is on your side, but it's subtle: the test that dies is not the one you updated. `test_update_advances_a_herdr_managed_checkout` stages its origin with `stage_origin` — no herdr-plugin.toml and no tags — so under your new rule the unversioned branch runs, finds zero release tags, and aborts. It passed before because the old fallback followed origin HEAD. Fix: give that fixture a readable version (or a release tag) so it exercises the normal in-major path it's actually about; your new `test_update_falls_back_loudly_without_a_readable_version` already covers the fallback.

Separately: please drop the 0.33.0 bump (all three files + the CHANGELOG heading) per the fork-PR rule in CLAUDE.md — the version is picked at release time on our side. Substantive review of the gates themselves to follow.

bartholomewtj:
Dropped the 0.33.0 bump (version files stay at 0.34.0) and merged current main.

`test_update_advances_a_herdr_managed_checkout` now tags 0.29.0 then 0.30.0 so it exercises the in-major dirty-lockfile path. The unversioned fallback stays in `test_update_falls_back_loudly_without_a_readable_version`.

AltanS:
Landed on main as 5f01bf7 + 4440c05 (cherry-picked with -x, your authorship kept) and released in v0.35.0 — thank you, the upload `__proto__` hole and the `.env` sourcing were real. I dropped the version bump/CHANGELOG from the fork commits (maintainer's call per CLAUDE.md) and wrote the 0.35.0 entry with a BREAKING block. Follow-up 9195e00: a failed `tailscale status` no longer bakes an empty allowlist into the unit (it kept locking the operator out), ARCHITECTURE §6 / DEPLOYMENT now match the fail-closed code, `.env` trailing `# comment` is stripped, and one composed test pins the shipped default.

PR #130, opened 2026-08-26
Fix forward-auth redirects after Authentik SSO expiry

lekoOwO:
## What

Recover cleanly when a fronting identity proxy answers an expired Collie API session with a redirect instead of a readable 401/403.

* force same-origin `/api/*` fetches to use `redirect: "manual"`
* translate intercepted redirects into a synthetic 401 so Collie's existing auth-error UI still works
* promote authentication to a top-level browser navigation instead of following the IdP redirect inside `fetch`
* detect Authentik's standard same-origin outpost via `/outpost.goauthentik.io/ping` and enter `/start?rd=...` directly
* fall back to Collie's existing operator-owned `/auth/` namespace for other proxies
* keep Authentik's fixed `/outpost.goauthentik.io/*` namespace out of the PWA navigation cache
* add regression tests for redirect interception and service-worker passthrough

## Why

Authentik's standard forward-auth setup commonly turns an unauthenticated request into a 302 redirect to `/outpost.goauthentik.io/start`.

For Collie's API polling, browser `fetch` follows that redirect by default. If the authentication flow leaves the app origin, the request can surface as a CORS/transport failure rather than the 401/403 that Collie's loader recognises as an authentication refusal.

The result is that an expired SSO session is reported as:

* `Reconnecting…`
* `Can't reach Collie`

instead of starting the authentication flow.

This change intercepts redirects from Collie's same-origin API requests before `fetch` follows them, feeds the existing auth-error path a readable 401, and starts authentication as a normal top-level navigation.

For Authentik, the standard outpost is detected through its `/ping` endpoint and recovery goes directly to:

```text
/outpost.goauthentik.io/start?rd=<current Collie URL>
```

For other identity proxies, Collie continues to use the existing `/auth/` escape hatch.

## PWA handling

Authentik's start and callback routes live under the fixed `/outpost.goauthentik.io/` namespace.

That namespace is added to the service worker's network-only navigation list so an installed Collie PWA cannot accidentally answer an Authentik start/callback navigation with the cached Collie app shell.

This is analogous to the existing `/cdn-cgi/` exception for Cloudflare Access.

## Scope

This does not make Collie manage Authentik or add a second managed front door.

The reverse proxy / identity provider remains operator-owned. Collie only recognises the redirect response shape and promotes authentication to a browser navigation.

## Tests

Added coverage for:

* same-origin `/api/*` detection
* manual 301/302/303/307/308 and `opaqueredirect` handling
* preserving pane `304 Not Modified`
* translating identity-proxy redirects into the existing 401 auth path
* leaving non-API fetches untouched
* Authentik `/ping` detection and `/start?rd=...` construction
* `/auth/` fallback when Authentik is absent or unreachable
* service-worker passthrough for Authentik start/callback routes

## Release metadata

Intentionally unchanged per the repository's fork-PR policy: no version bump or CHANGELOG entry in this PR.


AltanS:
Thanks — the diagnosis is right, and the `redirect: "manual"` → synthetic 401 half is exactly the fix we want: it brings back the Sign-in banner instead of "can't reach Collie". The SW denylist entry for the outpost prefix is fine too (same shape as `/cdn-cgi/`).

Two parts I can't take as-is:
1. **Auto-navigation.** `window.location.assign()` on any `/api/*` 3xx moves the page without a tap, with no visibility check and no loop guard (an IdP that bounces back unauthenticated ping-pongs forever). The banner's sign-in is a deliberate `<a>` (`connection-banner.tsx`) for that reason. Please feed the resolved entry into the banner's `href` instead.
2. **The Authentik `/ping` probe.** `api.ts` (the block above `XHR_HEADER`) states the stance: no proxy-specific branching, Collie manages nobody else's front door (ADR 0001). `DEPLOYMENT.md` already covers this case: redirect `/auth/` to your flow with one proxy rule.

Smaller: wrap `doReq` in `api.ts` rather than `globalThis.fetch`; add a `DEPLOYMENT.md` note; the new files trip v1's oxlint (`no-runtime-typeof`, the type-assertion rules in the test) — `main` merges into `v1` every release, so it has to pass there.

If you trim it to synthetic-401 + the SW entry, I'll merge it fast. No version bump / CHANGELOG in the PR please (fork rule in CLAUDE.md) — put the line in the description and I'll use it.

lekoOwO:
Thanks — updated. I removed the automatic navigation and Authentik probe, moved the redirect handling into the API client, and kept /auth/ as the user-initiated sign-in path.
I also added the deployment note and kept the Authentik outpost SW passthrough. No version/CHANGELOG changes in the fork PR.

PR #131, opened 2026-08-26
feat(quick-dock): let the operator declare their own quick replies

fucx:
The Quick dock is the last operator surface with a hardcoded list. The palette got `commands.toml`, the Keys tray got `keys.toml` — the dock still ships fixed English phrases. `quick-replies.ts` says as much in its own header: that list is content policy, not detection.

So an operator working in another language has no route to it, and neither does one whose harness would rather be told `approve` than `yes`. I hit the first case: my agents read German, the dock did not.

This adds `quick-replies.toml` on the same contract as its two siblings: same reader (`operator-file.ts`), same scope ladder (`operator-scope.ts`), replace rather than merge per ADR 0018, live behind the mtime check.

```toml
[[replies]]
scope = "claude"          # optional; omit for every pane
title = "bestätigen"
items = ["ja", "nein"]
```

One deliberate decision worth your eye: **a shell pane is addressable**, via `scope = "shell"`. An unscoped row reaches it too. The shipped `y`/`n` is the right default, but an operator replacing the dock in another language means that pair as much as the rest, and a shell that silently stayed English would be the one pane their file could not reach. Easy to argue the other way; say so and I'll change it.

Without the file, the `/api/config` payload and the rendered dock are byte-identical to before.

I left `CHANGELOG.md` alone: `check-version.sh` ties the newest heading to a version number, and that bump is yours to make.

Eleven tests cover the bridge grammar. A bad scope drops its row rather than widening it to every pane, one malformed row costs the file none of its other groups, a redefined title wins in place, and when the file stops parsing the reader keeps serving the last rows that worked. Five more cover the client resolution. The full suites still pass (672 bridge, 3434 web), as do both typechecks and a production build.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,6 +57,11 @@ COLLIE_SUBMIT_KEYS=Enter
 # Likewise your own preset chords: `keys.toml` next to this file, from `keys.toml.example`
 # (README → Your own key presets).
 
+# --- Quick-dock replies ---
+# And your own one-tap phrases: `quick-replies.toml` next to this file, from
+# `quick-replies.toml.example` (README → Your own quick replies). This is also where you replace
+# the shipped English phrases with another language.
+
 # --- Pane history (the agent's own session log) ---
 # On by default. This is the ONLY scrollback most agent panes can have: an agent TUI runs on the
 # terminal's alternate screen, which keeps no scrollback ring, so history is read from the log the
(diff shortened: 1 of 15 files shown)


fucx:
Per CLAUDE.md → Versioning, the CHANGELOG line in my words, for whenever you cut the release:

```
- **`quick-replies.toml`: your own Quick-dock groups** (title + items + optional `scope`), live-reloaded, replacing the shipped phrases on the panes they address per ADR 0018, shell panes reachable via `scope = "shell"` (9a4ad83)
```

Under `### Added`, and the hash is mine — swap it for the cherry-picked one.

AltanS:
Landed on main as eb1e92f (cherry-picked with -x, your authorship kept) and released in v0.35.0 — thank you! A small follow-up (0eb6d8d) taught the docs/module headers about the third sibling and added the parallel test. On `scope = "shell"`: yes, it stays addressable — it opens an option rather than closing one, no ADR needed.

PR #132, opened 2026-08-26
Fix Codex CLI first-class harness adapter

memset0:
The original implementation of Codex CLI first-class adapter lacks compatibility, especially when user is using a customized status bar configuration. This is a fix of this issue. I also extends the Codex adapter to recognize customized composers, working/queued drafts, large inputs, stable wrapped prompts, and exact slash-command palettes while preserving Collie's fail-closed dialog safety.

To summarize the problem, when someone try to send message to a codex agent with **Collie 0.33.0+** and customized status bar, the message wouldn't be send and Collie would report such error:

> Message didn't reach the input box

The following content is generated by my agent and I hope would give you an overview of my PR.

---

## Original problem

Collie's Codex adapter identified a writable composer by looking for a tail `›` prompt followed by the default status line containing:

`Context N% left/used`

That assumption breaks when users customize `tui.status_line` and omit the `Context` field. The real composer remains visible, but `composerReady` and draft extraction fail, so
Collie may type the message and then permanently withhold Enter.

Several other valid Codex states had the same symptom:

- Compact two-field customized status lines were not recognized.
- While Codex was working, `tab to queue message` replaced the normal status line.
- Large input could appear as `[Pasted Content N chars]`.
- A large `pane.send_text` write could retain only its first 1,024 bytes.
- Wrapped input could reflow after the first successful read, invalidating prompt binding.
- Typing `/status`, `/fast`, or another slash command replaced the status line with the slash-command palette, making every slash command unverifiable.

### Customized Codex status lines

- Keeps the existing default `Context N% left/used` path.
- Accepts bounded customized status lines without assigning semantic meaning to their fields.
- Supports 3–12 bounded ` · `-separated fields.
- Supports compact two-field status lines only when they match Codex's strict ANSI renderer signature.
- Continues rejecting disabled, missing, malformed, overlong, torn, or transcript-like status rows.

### Working and queued drafts

- Recognizes the official `tab to queue message` footer while Codex is working.
- Recovers a half-written queued draft without including the footer.
- Preserves the native blank composer layout.
- Keeps request-user-input, wizard, notes, approval, and other dialogs outside the composer grammar.

### Composer presentation and draft recovery

- Treats the dim `Ask Codex to do anything` placeholder as an empty composer.
- Does not discard the same words when they are ordinary non-dim user input.
- Hides only an empty native composer.
- Keeps non-empty idle and working drafts visible for recovery and diagnosis.
- Removes only the located status/footer row from non-empty composers.
- Keeps native dialogs visible alongside their structured controls.

### Reliable guarded submit

- Splits Codex input into Unicode-safe chunks of at most 900 UTF-8 bytes.
- Never retries a partially delivered chunk sequence.
- Reports later chunk failures as partial delivery and withholds Enter.
- Accepts `[Pasted Content N chars]` only when `N` exactly matches the current message's Unicode scalar count.
- Gives Codex a longer bounded read-only verification window.
- Requires two consecutive identical verified prompt/draft tails before submit.
- Binds the final Enter to the complete stable prompt/draft region rather than only the first wrapped `›` row.
- Still emits at most one Enter and never retypes during verification.

### Exact slash-command palettes

- Recognizes a complete no-argument `/command` only when it exactly matches the first cyan/bold selected palette option.
- Supports commands such as `/status`, `/fast`, `/model`, `/compact`, `/new`, and `/help` through the same generic grammar.
- Keeps the palette visible as raw terminal output.
- Requires two stable palette reads and a final server-side prompt binding before Enter.
- Partial filters such as `/sta`, inline arguments, mismatched selections, renderer-free lookalikes, dialogs, and output below the palette remain fail-closed.

## Safety properties

This does not restore blind Enter submission.

The Codex adapter still requires independently bounded tail evidence before treating a screen as writable. A dialog transition, ambiguous status row, incomplete slash filter,
changed prompt region, or failed read causes Collie to withhold Enter.

Other harness adapters are unchanged.

## Verification

- Backend: 713 tests passed, plus lifecycle contracts.
- Web focused Codex/reply coverage: 867 tests passed, 6 existing todos.
- Complete Web suite: 3,499 tests passed, 18 existing todos.
- Both TypeScript configurations passed.
- Production PWA build passed.
- No version or CHANGELOG changes are included, following the repository's fork-PR policy.

### DIFF
--- a/HERDR_API.md
+++ b/HERDR_API.md
@@ -84,6 +84,10 @@ the socket assumptions behind the design in [`ARCHITECTURE.md`](./ARCHITECTURE.m
     first and wrong hypothesis.
   - A `\n` inside `text` is delivered as a real newline keypress, not as pasted content. What the TUI
     does with it (submit vs. insert) is the harness's choice, not something the paste framing hides.
+  - A 2026-08-26 Herdr 0.8.0 Codex probe sent 1,123 synthetic ASCII bytes in one RPC and the
+    composer retained exactly the first 1,024. Two ordered sub-1,024-byte RPCs retained all 1,123.
+    The Codex guarded path therefore transports one logical input as Unicode-safe chunks; it still
+    verifies the complete logical message before sending Enter and never retries a partial write.
 - **An ack means "herdr took the bytes", never "the TUI acted on them".** Both `send_text` and
   `send_keys` return before the target program has read, let alone rendered, anything. So a
   successful RPC pair is not evidence a reply was delivered — a focused TUI dialog can swallow the
--- a/bridge/server.test.ts
+++ b/bridge/server.test.ts
@@ -536,6 +536,25 @@ describe("pane write prompt binding", () => {
     expect(client.texts).toEqual([["w1:p1", "hello"]]);
   });
 
+  test("matching expected_prompt submits an existing Codex draft without retyping it", async () => {
+    const client = new FakePaneClient();
+    client.text = "some output\n› ship it please\n\n  model · project · Context 99% left";
+    const { audit } = auditEntries();
+    const res = await replyPane(
+      client as unknown as HerdrClient,
+      cfg(),
+      "w1:p1",
+      request({ text: "", submit: true, expected_prompt: "› ship it please" }),
+      audit,
+      null,
+      "default",
+    );
+
+    expect(res.status).toBe(200);
+    expect(client.texts).toEqual([]);
+    expect(client.keys).toEqual([["w1:p1", cfg().submitKeys]]);
+  });
+
   test("stale expected_prompt returns prompt_changed and sends no keys", async () => {
     const client = new FakePaneClient();
     client.text = "Command finished";
--- a/web/src/components/agent-chat.test.tsx
+++ b/web/src/components/agent-chat.test.tsx
@@ -250,6 +250,15 @@ const STATUS_TEXT = [
   "  ← for agents",
 ].join("\n");
 
+const CODEX_SLASH_TEXT = [
+  "Codex output",
+  "\x1b[48;2;61;64;64m                                                            \x1b[0m",
+  "\x1b[48;2;61;64;64m\x1b[1m›\x1b[0m\x1b[48;2;61;64;64m /status                                             \x1b[0m",
+  "\x1b[48;2;61;64;64m                                                            \x1b[0m",
+  "  \x1b[1m\x1b[38;5;6m/status  show current session configuration and token usage\x1b[0m",
+  "  /\x1b[1mstatus\x1b[0mline  \x1b[2mconfigure which items appear in the status line\x1b[0m",
+].join("\n");
+
 // A minimal multi-question wizard tail (stepper header + current question) — enough for the REAL
 // wizard detector to lift it into the native WizardBlock inside AgentChat's mirror.
 const WIZARD_TEXT = [
@@ -419,6 +428,13 @@ describe("AgentChat — block-grammar scoping (an agent with no adapter)", () =>
     expect(strip.textContent).toBe("Local Llama (xhigh) · plan");
     expect(screen.queryByText(/Shift\+Tab:mode/)).toBeNull();
     expect(screen.queryByText("╭")).toBeNull();
+  });
+
+  it("keeps an exact Codex slash palette raw while treating its command as recoverable input", () => {
+    const codexAgent = { ...opencodeAgent, agent: "codex", paneId: "w8:p4" };
+    renderChat({ text: CODEX_SLASH_TEXT, agent: codexAgent });
+    expect(screen.getByText(/show current session configuration/).closest("pre")).not.toBeNull();
+    expect(screen.getByText(/configure which items appear/).closest("pre")).not.toBeNull();
   });
 });
 
--- a/web/src/components/composer.test.tsx
+++ b/web/src/components/composer.test.tsx
@@ -1445,6 +1445,59 @@ describe("Composer — terminal-draft preview", () => {
     await waitFor(() => expect(box).toHaveValue("")); // cleared after send
   });
 
+  it("Codex working draft Take over clears the queue composer before retyping and queues once", async () => {
+    const user = userEvent.setup();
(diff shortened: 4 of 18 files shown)


jxstanford:
Live-testing follow-up on `d57b9a5`: the original custom-status fix works, but there is one more Codex composer state that still triggers “The agent's input box isn't on screen”.

While a turn is working/steerable, Codex can render the composer with a purple `» ` marker instead of the idle `› ` marker. A sanitized version of the observed tail is:

```text
• Waiting for background terminal (2m 24s • esc to interrupt)

» Ask Codex to do anything

  gpt-example · ~/demo-project · Main [default]  Goal paused (/goal resume)
```

The status row is accepted by this PR, but `PROMPT = /^› (.*)$/` rejects the `» ` row, so `locateComposer` returns `null` and guarded reply preflight refuses the send. On the identical live capture, changing only `»` to `›` changes `composerReady` from `false` to `true`.

I tested this minimal local follow-up:

```ts
const PROMPT = /^[›»] (.*)$/;
```

The regression test uses the styled `»` marker and dim placeholder, and asserts:

- `composerReady === true`
- the placeholder remains an empty draft
- `composerPrompt === "» Ask Codex to do anything"` so the expected-prompt guard stays bound to the exact live row

Verification on top of this PR:

- RED before the parser change: the new test failed because `composerReady` returned `false`
- focused Codex adapter: 393 passed, 5 todo
- related composer/chat/nav tests: 155 passed
- full frontend suite with capped workers: 3,500 passed, 18 todo across 118 files
- root and web TypeScript checks passed
- production PWA build passed
- the real affected pane now returns `composerReady: true` with the exact `» Ask Codex to do anything` prompt signature

The existing tail status-row anchor and dialog safeguards are unchanged; this only widens the recognized column-zero composer glyph.

memset0:
> Live-testing follow-up on `d57b9a5`: the original custom-status fix works, but there is one more Codex composer state that still triggers “The agent's input box isn't on screen”.
> 
> While a turn is working/steerable, Codex can render the composer with a purple `» ` marker instead of the idle `› ` marker. A sanitized version of the observed tail is:
> 
> ```
> • Waiting for background terminal (2m 24s • esc to interrupt)
> 
> » Ask Codex to do anything
> 
>   gpt-example · ~/demo-project · Main [default]  Goal paused (/goal resume)
> ```
> 
> The status row is accepted by this PR, but `PROMPT = /^› (.*)$/` rejects the `» ` row, so `locateComposer` returns `null` and guarded reply preflight refuses the send. On the identical live capture, changing only `»` to `›` changes `composerReady` from `false` to `true`.
> 
> I tested this minimal local follow-up:
> 
> ```ts
> const PROMPT = /^[›»] (.*)$/;
> ```
> 
> The regression test uses the styled `»` marker and dim placeholder, and asserts:
> 
> * `composerReady === true`
> * the placeholder remains an empty draft
> * `composerPrompt === "» Ask Codex to do anything"` so the expected-prompt guard stays bound to the exact live row
> 
> Verification on top of this PR:
> 
> * RED before the parser change: the new test failed because `composerReady` returned `false`
> * focused Codex adapter: 393 passed, 5 todo
> * related composer/chat/nav tests: 155 passed
> * full frontend suite with capped workers: 3,500 passed, 18 todo across 118 files
> * root and web TypeScript checks passed
> * production PWA build passed
> * the real affected pane now returns `composerReady: true` with the exact `» Ask Codex to do anything` prompt signature
> 
> The existing tail status-row anchor and dialog safeguards are unchanged; this only widens the recognized column-zero composer glyph.

Thank you for feedback. Let me test it

AltanS:
Thanks @memset0. This PR did real work, and four pieces of it are on `main` today under your authorship, cherry-picked from `d57b9a5` and shipped in v0.36.0:

- `1ca57f1`: `codexDraftCarriesSend`, the `[Pasted Content N chars]` count as the evidence Enter waits for. Clean ADR 0010 work.
- `c58ba36`: the dim placeholder is empty; the same words typed are a draft.
- `47410a1`: `composerPrompt` binds to the whole wrapped draft run, so a long message no longer 409s every sweep.
- `ff412fc`: your bridge test for a prompt-bound submit.

The rest did not land. I want to be precise about why, because this is restructuring, not rejection.

**Chunking.** ADR 0010 forbids splitting a send into sub-threshold chunks by name, and the reason is concrete: `pane.send_text` has no bracketed paste, so a chunk boundary that lands on a lone `\n` submits a half-written message. `codexInputChunks` splits on byte count alone; I reproduced that case. Your 1,024 observation was right about the number and wrong about the layer: we live-probed `pane.send_text` on herdr 0.8.2 and 40,000 bytes arrived whole at the PTY (`HERDR_API.md` has the table), while a 3,000-character send into the Codex composer became `[Pasted Content 1024 chars]`. So the cut is Codex's own input handling. It is recorded as a known limit in the 0.36.0 notes, and the placeholder path you wrote is the right shape of answer to it.

**`isCustomStatusRow`.** A text-only "3 to 12 dotted fields" rule accepts ordinary indented prose, and it sits on the path that authorises destructive writes. Your `isStyledTwoFieldStatusRow` was the right idea: identify the widget by its renderer paint, not its text. That is what shipped, generalised to N fields and anchored on five byte-faithful 0.150.1 captures (`0ed3fd5`). It turned out to matter more than either of us knew: 0.150.1's default status line has no `Context` field at all, so the styled acceptor is what makes a plain 0.150.1 pane work.

**The `»` working composer, the queue hint, the slash palette, and the presentation changes** (`presentChrome`, dialogs rendering twice) each need their own PR, their own captured fixtures, and the Tier-2 bar in `HARNESS_CONTRIBUTING.md`.

@jxstanford, you saw the `»` marker live. We could not capture it here (no Codex subscription for a model turn). If you can run `scripts/capture-fixture.sh <paneId> codex--v0150-working` on a pane in that state and attach the file, that unblocks the mid-turn reply fix. It is listed as a known limit in 0.36.0 until then.

Closing this one. If you want to pick up the `»` composer or the slash palette as a fresh PR against 0.36.0, the fixture corpus is in place to build on.

PR #135, opened 2026-08-28
Worktrees from the phone: three declared capabilities

### LINKED ISSUE #133: Create a Git worktree from the phone
Right now, starting fresh work from the phone means starting it in whatever checkout the pane already sits in. If I want to branch off and put an agent on it, I have to walk back to the desk — which is the one thing Collie otherwise saves me from.

Herdr itself can already create a worktree and open it as a workspace; that capability just isn't reachable from Collie. So this doesn't look like it needs anything new from herdr — only a route to something that already exists.

**What I'd want, smallest useful version:** a "new worktree" action on a space that asks for a branch name, creates it, and drops me straight into a pane there, ready to send the first prompt. Browsing and removing existing worktrees would be nice later, but creation is the part that's actually blocking.

If you're happy with the idea, I'm glad to put up a PR.

broven:
Closes #133.

Branch off and put an agent on it from the phone. Three capabilities — `listWorktrees`, `createWorktree`, `openWorktree` — behind three routes, surfacing in two places.

### A worktree nests under the space showing its repo

spaces list with two worktrees nested under their repo

`MuxSpace` gains `repoRoot` and `isWorktree` — declared **facts**, like `spaces` capacity, never capabilities — and Herdr answers both on the workspace record it already sends, so the nesting costs no extra call and needs nothing switched on.

Three rules, each answering a case the flat list never had:

- **A group takes the position of its most recent member.** The list promises that what is fresh is near the top; sorting by the parent alone would bury a worktree used a minute ago under a checkout last touched last week.
- **A worktree whose repo is not open stays flat.** There is no row to indent under, and indenting under nothing reads as a rendering fault.
- **Filtering flattens**, for the same reason: a filter matching only the child would indent it under a parent that is not on screen.

A space outside any repo is the row it always was.

### `+` asks which kind

the new-space sheet's worktree tab: a repository picker and a branch name field

The worktree side asks **which repo**, because a sheet opened from the list has no current space to take one from. It offers the repos that are open, defaulting to the most recently used. Under the branch field sits *"or open one that already exists"* — the worktrees of the chosen repo that no space is showing, read once per repo choice. Those are the only checkouts the phone has no other route to; everything else is already a space in the list behind the sheet.

The tab strip renders **only where there is a choice**. No capability, or nothing open in a repo, and this is the plain new-space sheet, unchanged.

Opening navigates the phone. `focus` is never sent, so the operator's own screen stays where they left it — moving that is `setFocus`, which is its own capability for exactly this reason.

Both states are cards in the states playground, each with its "reach it for real" line.

### Why the port, and not `git worktree add` + `createSpace`

The cheaper route is the obvious one, and [ADR 0032](https://github.com/AltanS/collie/blob/v1/.adr/0032-a-worktree-is-opened-by-the-multiplexer-not-by-git.md) is there for when it is proposed again.

Herdr does not merely run Git; it keeps a record — `open_workspace_id` per checkout, a `worktree` block on every workspace. That record **is** the feature: it is how the phone knows which checkouts exist, which of them a space already shows, and which space nests under which repo. A checkout made behind Herdr's back has none of it, and Collie could only list it by doing its own Git work and then reconciling two answers about one directory.

Herdr declares all three. tmux and zellij decline them with their reasons: tmux could shell out to Git, but it keeps no mapping from a checkout to the session showing it, so what it made could not be found again. Declining is the honest answer until an adapter keeps that record itself and a probe proves it.

### Probed, not assumed

Every Herdr cell in `MUX_CONTRACT.md` cites a first-hand probe of **herdr 0.8.2 (protocol 20), 2026-08-28**, run against an isolated `herdr --session` over a throwaway repo — never a live pane. Two findings shaped the code:

- **`worktree.create` is not atomic.** In a session with no window server the checkout was created and the open failed with `worktree_open_failed`: the branch exists, nothing shows it, and creating again answers `worktree_create_failed` because the path is taken. That case gets its own code — `worktree.created_not_opened` — and the UI offers **open**, never **create again**.
- **`worktree.open` is idempotent**, answering `already_open` with the space showing it. Treated as an answer rather than a refusal: either way the pane below is where to go.

Refusals reach the phone as catalogued codes. Three change what it does — a busy multiplexer means retry, an ambiguous branch means type a better one, a space outside a repo shows nothing at all — and the rest carry the multiplexer's own sentence in `{reason}`.

### Notes

- **Version files and CHANGELOG untouched**, per CLAUDE.md's fork rule. A line in my words, if you want one:
  `- Worktrees from the phone: create or open a Git worktree from the spaces list, and see it nested under its repo. (<hash>)`
- Six locales, since the dictionary is total by type.
- Conformance covers all three in both directions. The declared side creates its own target first — "open this" cannot be proven against a checkout nothing made.
- `bun run lint` clean, `bun test` and the web suite green.


broven:
Marking this a draft while I run it on my own herd — please don't merge yet. I'll flip it back to ready once I've stopped finding things.

Two fixes already came out of that testing and are pushed: the spaces list was flattening away the repo→worktree relationship (four of my six spaces are two repos and their worktrees, and nothing said so), and creating one had no entry point from the list itself. Both are in the last commit.

One limitation worth naming, since it surprised me on a real herd rather than a probe: I have four worktrees on one repo and only two are open as spaces — and the two that aren't cannot be removed from the phone at all, because Herdr's removal is addressed by workspace. The sheet says so on the row and offers Open first, but it's a sharper edge in practice than it looked in the matrix.

broven:
Scope cut, and the PR is smaller for it: **removal is out**. What's left is list, create and open — three capabilities, three routes.

Not because it was hard. Because it was reaching past a decision Collie has already made: there is no `closeSpace` capability here, and Herdr's `worktree.remove` closes the space along with the checkout. Shipping it would have handed Collie its first space-destroying verb through a side door nobody argued for. A checkout that has outlived its use still gets cleaned up at the desk, which is where every other space-ending decision is made today.

`MuxWorktree.openSpaceId` stays, since it is what tells the phone which checkouts are already spaces — both the nesting and the "or open one that already exists" list are built on it. Removal can be added later against that same field, as its own argument.

ADR 0032 is reframed to match: the bookkeeping case for asking the multiplexer rather than running Git now rests on listing and re-opening, which is where it always did its real work. Removal still appears there — as the sharpest illustration of why a checkout made behind Herdr's back is one nothing can clean up — but as an illustration, not a verb being shipped.

AltanS:
Just a heads up, this PR is against the v1 branch (not main) and I dont  have a clear release path for v1 yet. Not sure if that was intended, but it definitely saves me some time as I won't have to port the feature back into v1. There is a path to install the beta which are outlined in the v1's readme. Let me know if you have troubles setting it up.

If this was meant for the current main branch, please close and resubmit the PR.

AltanS:
Merged onto v1, shipping in 1.0.0-beta.48. Thank you for the full stack, all seven of your commits landed with your authorship intact. Two adaptations on top: worktree creation now runs through the space-create flow (so the app navigates and refreshes like every other create), and the sheet was aligned with the v1 design rules. Closes #133 as well. Closing since the work is on v1.

PR #136, opened 2026-08-28
fix(ctl): find the Tailscale CLI inside macOS's app bundle

narze:
## The bug

On macOS, `collie-ctl.sh serve` (and so `start` / `restart`) reports:

```
error: tailscale not found; cannot publish the tailnet front door
note: the tailnet front door did not come up; the bridge is still on 127.0.0.1:8787
    tailnet   http://127.0.0.1:8787 (Tailscale name unavailable)
```

…on a Mac that is fully logged in and connected. The App Store and standalone Tailscale builds ship the CLI *inside* the app bundle, at `/Applications/Tailscale.app/Contents/MacOS/Tailscale`, and put nothing on PATH unless you ran the GUI's "Install CLI" step. Every `command -v tailscale` guard fails and the tailnet front door silently never comes up.

## The fix

Resolve the CLI once at load, the same way `resolve_bun` already resolves Bun: PATH first, then the usual absolute install locations, including the two app-bundle paths. Call sites use `"$TAILSCALE"`; the presence checks become `have_tailscale`.

The bundle directory is deliberately **not** prepended to PATH — the binary there is `Tailscale`, which only answers to `tailscale` on a case-insensitive volume, and the directory holds the GUI app too.

`COLLIE_TAILSCALE_BIN` overrides the result. Set to the empty string it makes Collie behave as if the host has no CLI, which is how `test_missing_tailscale_cli` now stages that condition — a bare PATH no longer hides an installed `Tailscale.app` from the resolver. `test_tailnet_acl_warning`'s `stage_tailscale` repoints `$TAILSCALE` for the same reason, extending the note already there about why that stub must be a real executable.

## Verified

`scripts/collie-ctl.test.sh` passes. On the reporting Mac (Tailscale 1.102.3, App Store build, no CLI on PATH):

```
tailscale serve (https) → tailnet :443 -> 127.0.0.1:8787
  ✓ Collie is running
    local     http://127.0.0.1:8787
    tailnet   https://<node>.<tailnet>.ts.net
```

Per CLAUDE.md this fork PR carries no version bump and no CHANGELOG entry. Suggested line if you want one:

> Fixed: macOS — find the Tailscale CLI inside `Tailscale.app` when it is not on PATH, so the tailnet front door comes up. (`COLLIE_TAILSCALE_BIN` overrides.)

https://claude.ai/code/session_01Mw6tyDHvBPtZ69Cth423bo

### DIFF
--- a/scripts/collie-ctl.sh
+++ b/scripts/collie-ctl.sh
@@ -184,6 +184,50 @@ case "$BUN" in
     esac
     ;;
 esac
+# Find the Tailscale CLI on PATH, then in the usual install locations.
+#
+# macOS is why this exists. The App Store and standalone Tailscale builds ship the CLI *inside* the
+# app bundle, named `Tailscale` (capital T), and put nothing on PATH unless you ran the GUI's
+# "Install CLI" step. So `serve` reported "tailscale not found" on a Mac that was fully logged in
+# and connected, and the tailnet front door silently never came up — the bridge stayed on
+# 127.0.0.1 with no hint that the CLI was two directories away.
+#
+# The bundle path is NOT put on PATH: the binary there is `Tailscale`, which only resolves as
+# `tailscale` on a case-insensitive volume, and the directory holds the GUI app too. Call sites use
+# "$TAILSCALE" instead, the same way they already use "$BUN".
+#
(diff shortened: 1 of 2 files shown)


narze:
Sorry, this was automated and I didn't expect it to open PR

PR #140, opened 2026-08-29
fix(codex): locate a composer whose continuation row is indented deeper than the gutter

sdertli:
## The symptom

Every reply sent from the phone to a **Codex** pane was refused with

> The agent's input box isn't on screen — a menu or dialog is probably up. Nothing was typed.

with no dialog anywhere on the screen. Claude, Grok and agy panes on the same host were fine.

It is a **deadlock, not a transient**: the refusal leaves the text in the composer, and the
stranded draft is what makes the *next* send fail too. The pane never recovers on its own — the
operator has to notice, clear the box by hand, and retype.

## The cause

`codex/chrome.ts` recognised a wrapped-draft continuation row with `CONTINUATION = /^ {2}\S/` —
"exactly two spaces, then a non-space". But the two spaces are the composer's **gutter**; what
follows them is the **operator's own text**, and that text may itself begin with spaces. One
shift+enter — a single tap on a phone keyboard — with the next line starting indented is enough:

```
› please move all the images across to the new blog
    then take the originals down once the copy is verified     ← four spaces
  gpt-5.6-sol medium · /tmp/collie-codex-sandbox
```

`locateComposer` walks up from the status row, hits that row, matches neither `promptText` nor
`CONTINUATION`, and returns `null`. `composerReady` is then `false` on every frame for as long as
the draft sits there, and `preflight` refuses the send.

## The fix

`CONTINUATION = /^ {2}\s*\S/` — assert the gutter, not the shape of what follows it.

Nothing else is loosened, and the discrimination this row's regex looked like it was doing is
still done, by the two checks that were always the load-bearing ones:

- the walk starts only from a row `isStatusRow` accepts at the buffer tail (dialogs replace the
  prompt/status pair, so their footer is the tail and the walk never starts);
- the run is bounded **above by the blank row**, tested first in the same expression — Codex
  separates every section of a screen with exactly one blank row, so a contiguous non-blank run
  directly under a validated status row is the draft by construction.

`› ` and `• ` rows still start at column 0, so neither can pass as a continuation.

## Evidence

`web/src/fixtures/panes/codex--v0151-draft-indented-line.txt` — a byte-faithful `format:ansi`
capture of the screen above from a throwaway herdr tab (Codex 0.151.0, herdr 0.8.2, Linux),
one length-preserving sanitization pass on the shell prompt line. It returns `composerReady:
false` on `main` and `true` with this patch; both were run before and after the change.

## Tests

`cd web && bun run test` → 118 files, 3841 passed / 30 todo.
`bun run test` (root) → 717 passed + `collie-ctl.test.sh` passed.
`bun run typecheck` on both sides → clean. `scripts/check-version.sh` → ✓ (0.36.0, untouched).

Per CLAUDE.md the version files and CHANGELOG are left alone — this is a fork PR. Suggested
CHANGELOG line, if you want one in my words:

```
- Codex: a draft whose continuation row is indented deeper than the gutter no longer reads as a dialog, which had refused every send in that pane until the box was cleared by hand
```

### DIFF
--- a/web/src/fixtures/panes/README.md
+++ b/web/src/fixtures/panes/README.md
@@ -69,6 +69,20 @@ model turn could be run.
 | `codex--v0150-custom-status.txt` | `-c 'tui.status_line=["model-with-reasoning","current-dir","git-branch"]'` (Context deliberately omitted) with a short draft on the `› ` row. Pins the styled custom-status design: per-field colours and dim ` · ` separators | `idle` |
 | `codex--v0150-paste-placeholder.txt` | One `pane.send_text` of exactly 3000 non-newline ASCII characters lands as `[Pasted Content 1024 chars]` — **N is 1024, not 3000**. Codex's TUI keeps only the first 1024 characters of a single burst, so `draftCarriesSend("x".repeat(3000), draft)` is **false** (it is true for a 1024-character send). See `codex/paste.ts` | `idle` |
 
+## Codex 0.151.0 capture (2026-08-29, herdr 0.8.2, Linux sandbox pane)
+
+Byte-faithful `format:ansi` capture from a throwaway herdr tab in `/tmp/collie-codex-sandbox`, with
+one length-preserving sanitization pass on the shell prompt the pane opened with (`user@sbox`, `$`).
+Status row is 0.150.1's two-field default, so the styled acceptor is what locates the composer here
+too. **The headline: a continuation row is NOT always two-space indented.** The gutter is two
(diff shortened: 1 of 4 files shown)


AltanS:
Thank you — the root cause here is a good catch: treating the gutter's two spaces as the boundary when the operator's own text can start with spaces. The fixture you added is what makes it stick. Verified against `main` with #141, #142 and #121: no conflicts, full suite green.

PR #141, opened 2026-08-29
fix(codex): recognize dim status suffix

mauhai:
## The symptom

Every reply sent from Collie to a current Codex pane was refused with:

> The agent's input box isn't on screen — a menu or dialog is probably up. Nothing was typed.

The composer was visibly present and held an ordinary one-line draft (`test`); no dialog was open.

## The cause

Current Codex paints the final collaboration-mode status field together with its separator as one
dim ANSI segment:

```text
<coloured model> · <coloured cwd><dim> · Main [default]</dim>
```

Collie's context-less status-row detector accepted only alternating coloured fields and standalone
dim separators. It therefore rejected this valid status row, so `locateComposer` returned `null`
and the guarded reply path refused to type.

This is separate from #140: that PR fixes deeper indentation in multiline draft continuation rows;
this failure occurs with a one-line draft and is in the status-row parser.

## The fix

Accept one combined dim `separator + field` suffix only when it follows at least two ordinary
coloured fields and is the final segment. The existing paint checks, field bounds, control-character
rejection, and tail-anchored prompt/status shape remain intact. Unstyled lookalikes still fail closed.

## Evidence

The parser was run directly against the affected live pane's `format:ansi` output after the change:

```json
{"composerReady":true,"draft":"test","status":["  gpt-5.6-sol medium · ~/.herdr/worktrees/kausate/feature-kau-6102-repair-dex-services-and-run-tabs · Main [default]"]}
```

The rebuilt Collie page was reloaded against that pane; the false refusal disappeared and the draft
remained ready to send.

## Tests

- pre-push backend typecheck and web typecheck: passed
- backend suite: 691 passed
- `collie-ctl.test.sh`: passed
- web suite: 3,827 passed / 30 todo
- production build: passed
- `scripts/check-version.sh`: passed at 0.36.0

Per `CLAUDE.md`, version files and `CHANGELOG.md` remain untouched because this is a fork PR.

Suggested changelog line:

```text
- Codex replies work with the dim collaboration-mode status suffix — the composer gate recognises the final ` · Main [default]` renderer segment without accepting unstyled lookalikes
```

### DIFF
--- a/web/src/lib/harness/codex.test.ts
+++ b/web/src/lib/harness/codex.test.ts
@@ -297,6 +297,26 @@ describe("the styled status-row acceptor fails closed", () => {
     expect(isStatusRow(text, line)).toBe(true);
   });
 
+  it("accepts Codex's dim final status field", () => {
+    // Current Codex paints the final collaboration-mode field together with its separator:
+    // `...<coloured cwd><dim> · Main [default]</dim>`. This is the live shape that left Collie's
+    // composer visible but made the reply pre-flight report that no input box was on screen.
+    const { text, line } = row(
+      `  ${FG}gpt-5.6-sol medium${OFF}${SEP}${FG2}/tmp/project${OFF}${DIM} · Main [default]${OFF}`,
+    );
+    expect(isStatusRow(text, line)).toBe(true);
+  });
+
+  it("accepts the dim suffix only after two painted fields and only at the end", () => {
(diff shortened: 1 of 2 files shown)


AltanS:
Thank you — this is a well-diagnosed fix, and the separation from #140 in your description made it easy to review. Verified against `main` alongside #140, #142 and #121: no conflicts, full suite green (717 bridge tests, 3866 web tests, both typechecks, the ctl lifecycle suite).

PR #144, opened 2026-08-31
fix(codex): clean up mobile terminal chrome

foreverrrree:
## Summary

- keep Codex's labelled, terminal-width rule rows on one visual line in the wrapped mobile mirror
- suppress Codex's observed `rgb(240,240,240)` submitted-message fill at phone width, where the light-theme mirror inversion turns it into a solid black bar
- preserve the native ANSI background on desktop and leave semantic diff backgrounds untouched

## Why

Codex pads submitted user-message rows to the terminal width and paints the whole row near-white. Collie's light-theme mirror inversion maps that fill to near-black, producing a heavy full-width bar on a phone. Codex also emits labelled separators such as `─ Worked for … ───` that do not satisfy the global pure-rule classifier and therefore wrap into several visual rows.

Both exceptions are presentation-only and scoped to the exact `codex` adapter. Other harnesses retain the existing global wrapping and colour contracts.

## Testing

- `cd web && bun run test -- --reporter=dot --maxWorkers=2 --testTimeout=15000` — 118 files, 3,870 passed, 30 todo
- `bun run build`
- `bun test ./bridge ./scripts` — 717 passed
- manually verified against a live Codex pane at a 430px mobile viewport: submitted-message black bars are gone and labelled `Worked for` rules stay on one row

`scripts/collie-ctl.test.sh` was also attempted, but this Linux host's `plutil` rejected the generated macOS launchd fixture before reaching any frontend-related assertion.


AltanS:
Thanks for this. The black bar is real and the fix is in the right layer, so I want to take it. Three things first.

**The PR shows 519 files, but your change is one commit of 110 lines.** Your branch sits on the old 0.36.1 history. `main` was rewritten for the 1.0 cutover on 2026-09-01, so GitHub now compares against a merge base from July. Please rebase your one commit (`a71e25b`) onto the current `main`. I cherry-picked it locally: it applies cleanly, the web suite passes (161 files, 4717 tests), but `bun run lint` fails on two anti-slop rules:

```
web/src/lib/harness/codex/display.ts:30:7  no-conditional-empty-object-spread
web/src/components/ansi-output.test.tsx:122:22  require-safety-comment-for-type-assertion
```

Build the line object in plain statements instead of `...(noWrap ? {…} : {})`, and give the assertion a `SAFETY:` comment stating the checked invariant.

**The no-wrap rule is broader than the bug.** `LONG_CODEX_RULE_RUN` fires on any transcript line with 20 identical rule glyphs anywhere in it, dashes included. A no-wrap line is clipped on a phone, so a code block or table Codex prints could lose its right edge. Please match Codex's labelled separator shape instead, a row that starts with a short rule, a label, and a rule to the edge, so ordinary output cannot trip it.

**Please add a byte-faithful capture.** No Codex fixture in `web/src/fixtures/panes/` contains the `rgb(240,240,240)` fill or a labelled `─ Worked for … ───` row. The colour match is an exact string, so without a capture nothing pins it when Codex or the parser changes. One capture that shows both rows is enough; the corpus README explains the format.

Two smaller notes. The change to `blocks.test.ts` loosens a Codex-must-not-alter-lines assertion from identity to text equality. That is honest and fine, but say so in the commit message. And the CSS uses `!important` with a 640px media query that nothing else in the stylesheet uses; if the segment class can set the background directly, prefer that.

Leave version files and CHANGELOG alone, as you did. I write the Unreleased line on merge.

AltanS:
Thanks @foreverrrree. This is merged as 0104d27.

We cherry-picked your commit and kept it as its own commit on main
(74bf208), then fixed on top in c397622:

- LONG_CODEX_RULE_RUN fired on any row holding 20 identical rule glyphs
  anywhere, so a code block or a table could be clipped and lose its
  right edge. LABELLED_RULE_ROW replaces it and matches the whole row.
- The !important is gone. A marked segment carries its ANSI fill in
  --terminal-seg-bg, so the two class rules win by normal cascade order.
- The identity assertion in blocks.test.ts is restored;
  decorateCodexDisplay returns its input unchanged when no row matched.

One ask. The fixture
web/src/fixtures/panes/codex--submitted-fill-labelled-rule.txt was
reconstructed from your report, not captured from a real session. The
corpus README says so. If you still have the pane that showed this, a
real capture would be better than our reconstruction.

PR #146, opened 2026-09-01
fix(ctl): find the Tailscale CLI when Herdr's process PATH omits it

Fredddi43:
## The bug

On macOS, `collie-ctl.sh` (and so `herdr plugin action invoke start/restart/update`) fails to publish the tailnet front door on a fully logged-in, connected machine:

```
error: 'tailscale status' named no host for this node — the allowlist was not discovered.
       no allowlist is set, so the Host gate will refuse every request. Set
       COLLIE_TAILSCALE_HOSTS (or COLLIE_PUBLIC_HOSTS) in .env, or fix Tailscale and retry.
error: tailscale not found; cannot publish the tailnet front door
note: the tailnet front door did not come up; the bridge is still on 127.0.0.1:8787
```

Herdr spawns plugin actions with a minimal process environment — no login shell, so nothing has sourced whatever profile line put Tailscale on `PATH` (a Homebrew `bin` dir, or the macOS app bundle's CLI for anyone who never ran the GUI's "Install CLI" step). Every `command -v tailscale` guard in the script then reports "not found" even though Tailscale is installed, running, and reachable. The bridge keeps working locally on `127.0.0.1:8787`, it's just silently never exposed on the tailnet.

Root cause is identical to the one `resolve_bun` already works around for Bun (same file, just above): a bare `command -v <tool>` guard trusts the invoking process's `PATH`, and Herdr's plugin-action `PATH` isn't that.

## The fix

`resolve_tailscale()`, same shape as `resolve_bun()`: `PATH` first, then the usual absolute install locations, including the macOS app-bundle path. Every `tailscale`/`command -v tailscale` call site now goes through it (`"$(resolve_tailscale)"` / a new `have_tailscale` helper).

Two things worth calling out that came up while adding tests for it:

- **`COLLIE_TAILSCALE_BIN`** overrides the result outright, including to the empty string, which forces "no CLI" behavior. `test_missing_tailscale_cli` needs this: restricting `PATH` alone (the existing trick used for `bun`) doesn't defeat `resolve_tailscale`'s absolute-path fallbacks on a machine that actually has Tailscale installed at one of those fixed paths — so the "missing CLI" case would otherwise quietly stop reproducing on exactly the kind of machine this fix targets.
- **Not cached at load**, unlike `$BUN`. `test_tailnet_acl_warning` sources `collie-ctl.sh` once per harness run and then restages a different fake `tailscale` binary on `PATH` per case within that same process. A `TAILSCALE="$(resolve_tailscale)"` resolved once at source time (which is what I tried first) freezes on whatever was on `PATH` at that instant, so every case after the first silently tests stale state instead of its own fixture. Resolving fresh at each call site keeps that test's per-case restaging working, at the cost of a few extra `command -v`/`stat` calls per invocation — cheap for a script that isn't in a hot loop.

The app-bundle directory is deliberately never added to `PATH` — the binary there is named `Tailscale`, which would only resolve as `tailscale` on a case-insensitive volume, and the directory also holds the GUI app itself.

## Verified

- `scripts/collie-ctl.test.sh` passes in full.
- `scripts/check-version.sh` passes (no version files touched — this is a fork PR).
- Reproduced the original bug and confirmed the fix directly:
  ```
  $ env -i HOME="$HOME" PATH="/usr/bin:/bin" HERDR_PLUGIN_CONFIG_DIR=... bash -c '
      source scripts/collie-ctl.sh
      echo "resolved: $(resolve_tailscale)"'
  resolved: /opt/homebrew/bin/tailscale
  ```
  (On `main`, the equivalent `command -v tailscale` under that `PATH` finds nothing.)

## Re: #136

#136 proposed essentially the same fix and was closed unmerged — the author's comment says it was an unintended automated PR ("Sorry, this was automated and I didn't expect it to open PR"), not a technical rejection. I confirmed the bug it describes is still present on current `main` before writing this. Suggested CHANGELOG line, per `CLAUDE.md`'s fork-PR convention:

> Fixed: macOS/Linux — find the Tailscale CLI via absolute install paths (including the macOS app bundle) when it isn't on PATH, so the tailnet front door comes up under Herdr's minimal plugin-action environment. (`COLLIE_TAILSCALE_BIN` overrides.)


Fredddi43:
Addressed the Copilot review comment in 3059aaf: `self_dnsname` (called right after the `have_tailscale` guard in `cmd_serve`) shelled out to bare `bun` with no availability check, unlike its sibling `self_hosts`. Under `set -euo pipefail` a missing `bun` there was a raw "command not found" crash instead of a clean error.

Fixed by adding the same `command -v bun >/dev/null || return 0` guard `self_hosts` already has, so a missing `bun` makes `self_dnsname` return empty, and the existing empty-hostname handling in `cmd_serve`/`bridge_url` takes it from there. Verified tailscale-present/bun-absent now exits 1 with a clean "cannot determine Tailscale hostname" message instead of crashing. Full test suite still passes.

Fredddi43:
Closing — superseded by the 1.0.0 rewrite. The bash logic this PR patched (`collie-ctl.sh`'s `resolve_tailscale`/`self_dnsname`/`cmd_serve`) no longer exists; `scripts/collie-ctl.sh` is now a bootstrap shim that execs a compiled binary built from `cli/` + `bridge/`.

More importantly, the underlying bug is already fixed there, more generally: `bridge/tools.ts`'s `findTool()`/`fallbackDirs()` now resolves EVERY shelled-out tool (tailscale, git, systemctl, journalctl, herdr) through PATH-then-absolute-fallback, including `/opt/homebrew/bin` — and its own comment cites this exact bug. Verified by reading `bridge/tools.ts` on `main`.

Closing rather than rebasing, since there's no longer any bash code left for this diff to apply to.

One gap I noticed that isn't covered by the new `fallbackDirs()`: it doesn't include `/Applications/Tailscale.app/Contents/MacOS/Tailscale`, so a user with only the GUI app (never ran "Install CLI") would still hit "tailscale not found" — the same case #136 was originally about, just narrower now that the common Homebrew/usr-local case is fixed. Happy to send a small follow-up for that if useful.

PR #150, opened 2026-09-02
fix(omp): recognize OMP 18.1.2 composer format

### LINKED ISSUE #149: OMP 18.1.2 open-ended composer is not recognized by the reply guard
## Summary

Collie's OMP adapter does not recognize the composer format emitted by OMP 18.1.2 on a wide pane. The normal reply box reports that the agent input box is not on screen and asks for **Type anyway?**. If text is forced through, post-type verification also cannot find the draft, so Collie withholds Enter and reports that the message did not reach the input box.

## Environment

- Collie 1.1.0
- OMP 18.1.2
- Wide Herdr pane

## Observed shape

OMP 17's classic composer uses a closed bottom row and a separate top border:

```text
╭── <status> ──╮
╰─ <draft> ─╯
```

OMP 18.1.2 can emit a standalone status row followed by an open-ended prompt row:

```text
<standalone status row>
╰─ <draft>
```

The empty composer is a bare `╰─` row. Other OMP panels still close corner-to-corner (`╰────╯`).

## Root cause

`composerBottomText` only accepts `^╰─ (.*) ─╯$`, and `locateComposer` then requires an adjacent `╭…╮` top border. The OMP 18.1.2 shape therefore makes `locateComposer`, `composerReady`, and `extractInputDraft` all return no match.

## Expected behavior

Collie should recognize both OMP composer formats, extract the visible draft (including inline-suggestion handling), bind destructive pre-clear keys only when the composer is positively identified, and submit only after the typed draft is verified.

## Acceptance criteria

- Recognize OMP 18.1.2's open-ended `╰─ <draft>` row and its standalone status row.
- Preserve support for the closed OMP 17 composer.
- Continue declining modal/picker boxes that close corner-to-corner.
- Never send Enter without draft evidence.
- Add parser, adapter, and end-to-end composer regression coverage.

ImArtisann:
## Summary

- recognize OMP 18.1.2's open-ended `╰─ <draft>` prompt row alongside the classic closed `╰─ <draft> ─╯` form
- treat the standalone row above the open prompt as OMP's status chrome
- reuse the same draft and inline-suggestion extraction for both formats
- preserve the fail-closed modal rule: corner-to-corner picker/dialog boxes still do not qualify as composers

Fixes #149.

## Verification

- `bunx vitest run src/lib/harness/omp/markers.test.ts src/lib/harness/omp/chrome.test.ts src/lib/harness/omp.test.ts src/lib/reply-action.test.ts src/components/composer.test.tsx` — 717 passed, 9 existing todos
- `bun run typecheck`
- live smoke against an isolated OMP 18.1.2 pane: Collie's normal type box submitted `/context`, cleared the local draft, and OMP rendered the context report

### DIFF
--- a/web/src/lib/harness/omp/chrome.test.ts
+++ b/web/src/lib/harness/omp/chrome.test.ts
@@ -15,7 +15,7 @@ import {
 } from "./chrome";
 import { lineText } from "./markers";
 
-// omp's composer chrome. The whole adapter's Tier-1 value — and its safety half — is here: the box
+// omp's composer chrome. The whole adapter's Tier-1 value — and its safety half — is here: the shape
 // this scanner finds is the statusline, the stranded draft, AND the answer to "may a phone reply be
 // typed right now". A false "yes" types the user's message into whatever modal has the keyboard.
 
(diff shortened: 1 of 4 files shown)


AltanS:
Merged by rebase as 6081520, with your commit and authorship kept. It ships with the next release; the changelog line credits you and #149. Thank you for the careful root cause and the live smoke test.

PR #151, opened 2026-09-02
fix(web): omp panes fall back to an initials tile

enieuwy:
`AgentIcon` has no `omp` entry, so every oh-my-pi pane falls back to the neutral initials tile. Herdr reports `agent: "omp"` exactly (checked with `herdr pane list`); codex works only because `codex` is in the table — nothing about omp was special except its absence. The fold is a bare `omp` prefix, exactly as `canonicalAgent` already folds it for the slash catalog.

The mark is omp's own path from `omp.sh/favicon.svg`, rescaled ×0.6 and centred so its 64×64 geometry fills 24×24 (the tile adds its own padding) — the same path, not a redrawing. Rasterised against the original: shape IoU 100.0000%, 0 mismatched ink pixels of 320,240.

`grad` is new and optional, because omp's official mark is a three-stop gradient in every asset upstream publishes — the favicon, and the TUI's own welcome logo (`PI_LOGO`'s gradient stops in oh-my-pi's `welcome.ts`). It is **not** a contrast decision: flat `#9B4DFF` on `#0F0A14` measures 4.56:1 and reads fine at every size `AgentIcon` renders, down to the tab strip's `size-3.5`. Flattening would just ship a mark omp never paints, and land it a shade off the `pi` tile it sits beside in the same list.

The field costs one optional key and no new state: two stops minimum by type, so a gradient that paints nothing is unrepresentable; absent is how every existing brand keeps saying "flat"; and the gradient id is sanitised the way `collie-mark.tsx` already sanitises its own, so a column of tiles cannot collide on one id.

Also tightens the shared brand-table assertion, which used to accept `fill="none"` for any brand — a fill-mode mark that had lost its paint entirely still passed.

## Verification

- `bun run build` (root tsc + web tsc + build) green.
- `agent-icon` tests 22/22, 6 of them new. Mutation-checked the tightened assertion: forcing `fill="none"` fails 8 of 22.
- Live UI on this branch against real Herdr panes: every omp row shows the gradient π, 11 tiles with 11 unique gradient ids, every `url(#…)` resolving.

Version files and CHANGELOG deliberately untouched, per the fork rule in CLAUDE.md.

### DIFF
--- a/web/src/components/agent-icon-data.ts
+++ b/web/src/components/agent-icon-data.ts
@@ -2,14 +2,23 @@
 // official logo, paired with the brand's tile color. Sources (all verified against each project's
 // own favicon/site): Claude + Codex(OpenAI) via Simple Icons (CC0); pi via Simple Icons / pi.dev
 // favicon (#09090b tile); opencode via opencode.ai's site mark (#080808 tile, theme_color); agy
-// (Antigravity) via its Google-blue #1A73E8 tile, keyed under both "agy" and "antigravity".
+// (Antigravity) via its Google-blue #1A73E8 tile, keyed under both "agy" and "antigravity"; omp via
+// omp.sh/favicon.svg (#0f0a14 tile, and the one mark whose official paint is a gradient — `grad`).
 // To refresh: re-run the fetch in CHANGELOG and replace the `d` strings.
 
 export interface AgentBrand {
   /** Tile background — the agent's official brand color. */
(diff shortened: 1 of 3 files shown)


AltanS:
Merged by rebase as 17386ef, commit and authorship kept. The changelog line credits you. It ships with the next release. Thanks for the fidelity work on the gradient and the tightened brand-table check.

PR #153, opened 2026-09-02
feat: list live omp skills in the Agent-commands sheet

Giardi77:
## Why

The phone palette only shipped a static omp catalog. Per-project skills and extension commands (`/skill:…`, `/green`, `/review`) live on the running session, so they never showed up unless you spawned `omp --mode rpc` per pane — which discussion #117 sketched for pi, but is the wrong layer for omp.

## What

The **running omp** writes a hint file (not a control channel — ADR 0024, same as beacons). Collie reads it on each snapshot and **appends** those rows to the Agent-commands sheet.

- Shipped `/new` `/compact` stay on the first screen.
- Live extras are search-only (`common: false`).
- `commands.toml` still **replaces** the shipped catalog (ADR 0018); live extras still append after that list.
- Collie still types `/name` into the PTY. The dump is never obeyed as a control channel.

Install: copy `contrib/omp/collie-slash-catalog.ts` to `~/.omp/agent/extensions/` and **restart omp**. Dumps land in `~/.local/state/collie/slash-catalog/herdr-<paneId>.json`.

## Out of scope

Full TUI autocomplete (subcommands, `@file`, ghost accept). That still lives in omp's overlay.

## Version

From a fork: no version bump / CHANGELOG, per CONTRIBUTING.


Giardi77:
Closing at the author's request.

Issue #53, opened 2026-07-29
Phone mirror pans horizontally for most lines: wrap is off by default, and `source: "recent"` bakes in the desktop pane's column width


waynehoover:
## What happens

Reading Claude output on my phone, most lines run off the right edge and I have to pan
sideways to read them. Turning on **View → WrapText** helps, but the wrapping it produces
is ragged, because by then the lines have already been broken at the desktop's width.

## Why

Two separate things stack up.

**1. Wrap defaults to off.**
`DEFAULTS` in `web/src/hooks/use-display-prefs.ts` is `{ wrap: false, ... }`, and
`AnsiOutput`'s `wrap` prop defaults to `false`. With wrap off, `preClass()` renders
`whitespace-pre overflow-x-auto`, so the `<pre>` becomes a horizontal panner. The comment
says this is deliberate ("preserves column alignment like desktop Herdr"), which is the
right call for TUI tables. It is the wrong default for reading agent prose on a phone,
which is most of what the mirror shows.

**2. The bridge asks for the pre-wrapped snapshot.**
`bridge/server.ts:458` reads:

```ts
const read = await herdr.readPane(paneId, "recent", lines, "ansi");
```

`source: "recent"` returns the terminal grid, so every line is already hard-broken at the
*desktop* pane's column count. herdr spawns those panes at the desktop terminal's size; in
my session that is 190 columns (`spawning pane terminal ... rows=63 cols=190` in
`herdr-server.log`). A phone in portrait shows roughly 45 to 50 columns at the default
12px font.

Measured on one of my live panes, 200 lines, ANSI stripped:

| `source` | lines | longest line | lines over 60 cols |
|---|---|---|---|
| `recent` (used today) | 199 | 187 chars | 84 of 199 |
| `recent-unwrapped` | 176 | 721 chars | 79 of 176 |

So ~42% of lines need panning, and the 187-char maximum is exactly the desktop pty width.

The consequence is that turning wrap on does not really fix it: the browser re-wraps lines
that were *already* wrapped at 190 columns, so breaks land mid-sentence at arbitrary
points and Claude's own indentation and continuation structure is lost. You get correct
text in a shape that reads worse than either endpoint.

## Suggested direction

Both of these are small, and the second is the one that actually matters:

1. **Default `wrap` to `true`.** The pref is already persisted and already has a UI
   toggle, so anyone who wants column-faithful panning keeps it one tap away. `v3` is
[...]

Issue #156, opened 2026-09-02
Key path on phones: frame rows wrap into unreadable menus; poll/settle constants vs measured latency


alexlee2046:
Two small things from measuring the key path on a phone (herdr 0.8.2, iPhone PWA, solo install). Both are one-liners or near-one-liners; happy to split or adjust.

### 1. Frame rows should not wrap on a narrow mirror

Herdr spawns panes at desktop width (100–190 cols) while the phone mirror shows ~45–50, and the mirror's default is `wrap: true`. For prose that is right, but a box-drawing TUI menu (e.g. pi's `/model` picker) then wraps every frame row into 2–3 ragged visual lines: the frame scrambles and the inverse-video selection is shredded across rows, so the operator cannot tell which option the cursor is on — exactly while driving it with the Keys pad.

The existing `noWrap` path only covers **pure** horizontal rules. Extending it to frame **rows** (first and last glyph both frame glyphs) keeps each menu row on one visual line, so the frame stays aligned and the selection marker (at line start) stays readable; overflow past the mirror is clipped, which is fine since the marker lives left. Prose, `tree` output (leading `│` only, no trailing glyph) and normal text keep wrapping.

```diff
--- a/web/src/lib/blocks.ts
+++ b/web/src/lib/blocks.ts
@@
 const MIN_NO_WRAP_BORDER_LENGTH = 20;
 const PURE_HORIZONTAL_BORDER = new RegExp(
   `^([${PURE_HORIZONTAL_RULE_GLYPH_CLASS}])\\1{${MIN_NO_WRAP_BORDER_LENGTH - 1},}$`,
 );
+
+// Frame rows: both edges are frame glyphs (menu rows, panel borders). Wrapping scrambles the
+// frame and shreds the selection highlight; one visual row per line keeps the layout legible.
+const FRAME_EDGE_GLYPHS = "│┌└╔╚╠╟┐┘╗╝╢╡┤┏┗┓┛";
+const FRAME_ROW = new RegExp(`^\\s*[${FRAME_EDGE_GLYPHS}].*[${FRAME_EDGE_GLYPHS}]\\s*$`);
 
 function styledLine(segments: AnsiSegment[]): StyledLine {
   const text = segments.map((segment) => segment.text).join("");
-  return PURE_HORIZONTAL_BORDER.test(text.trim()) ? { segments, noWrap: true } : { segments };
+  return PURE_HORIZONTAL_BORDER.test(text.trim()) || FRAME_ROW.test(text)
[...]

PR #158, opened 2026-09-03
A table in the mirror pans instead of wrapping, and the prose around it still wraps

enieuwy:
Wrap lines is right for almost everything the mirror shows and wrong for one shape. A table's meaning is the column a character sits in, so reflowing it at phone width puts row three's second cell under row two's third: every row is still there and the table is unreadable. Turning Wrap off to read one table then costs column-faithful panning on all the prose too. This is the road issue #5 originally asked for ("no-wrap + horizontal scroll inside the mirror") and that #53 correctly overruled as a global default.

`lib/table-run.ts` groups a table's rows into one scroller inside the wrapping `<pre>`. Wrap stays on and there is no new setting.

### Before and after

Same idle pane, same 191-column table an agent printed, 390px viewport: shipped 1.2.0 on the left column of the description below, this branch on the right. On 1.2.0 the header row breaks so that `Channel` and `Evidence` share a line and `Current meaning` sits alone on the next, with the borders sprayed across six rows. On this branch the table keeps its columns and clips at the viewport edge, ready to pan; the prose above and below it still wraps. I can attach the two PNGs to a comment if you would like them in the thread.

Measured on the live pane: the run reports `scrollWidth 1128` in `clientWidth 374`, the outer `<pre>` reports `scrollWidth === clientWidth` (so the page itself never pans), and one `scrollLeft` moves every row of the table by the same amount.

### How it decides

One scroller per table, never one per row — two rows at different `scrollLeft` come apart exactly the way wrapping already breaks them.

Three grammars, each anchored on a row nothing else prints:

- a markdown delimiter row (`| --- | --- |`, with or without outer pipes),
- a `+---+---+` rule (mysql, sqlite3 `-table`, psql, docker, tabulate),
- a box frame row carrying a **cross**.

A T-piece is deliberately not enough, and that exclusion is load-bearing: a `┬` sits wherever any two-pane box's divider meets its lid, so an earlier revision that accepted one claimed omp's welcome splash and the whole of its `/model` picker across 18 of the 121 committed pane captures. The cost is a headerless box table, which keeps wrapping.

The anchor alone never claims a row. Markdown grows by pipe count; the two grid grammars grow by **column offset** — a member must carry a separator at each offset the anchor spends one on. That single rule is why a titled lid stays in the run, why a plain rule beside a table does not, why a chrome box below a table cannot be swallowed even with no blank line between them, and why the remains of a row the terminal already wrapped are left to wrap instead of being dragged in half-panned.

### Scope of the fix

Only the columns between the phone and the **pane** are recoverable. `pane.read` returns a rendered grid, so a table already too wide for the pane was broken by the terminal before Collie saw it, and no CSS gets those rows back (ADR 0008). Codex is worth noting here: its own TUI reflows a markdown table into stacked field lists at the pane's width, so there is never a table on the grid to claim, and a live codex pane correctly yields zero runs.

### Rendering details

The run is an `inline-block` sitting between the same `"\n"` text nodes as every other line, so find offsets, link offsets and a clipboard copy are untouched — a selection dragged across a run boundary serialises character-identically to `pre.textContent`. `overflow-y` is pinned explicitly, because `overflow-x-auto` alone computes it to `auto` and `LINK_CLASS`'s em-padding would then make the run a second vertical scroller. The run hides its own scrollbar: a classic non-overlay scrollbar is laid inside an auto-height box and would cover the last row with y pinned. It is keyed by its own first row rather than by a line index, so a poll that shifts the grid does not remount the scroller and throw away the reader's pan (verified live: a 200px pan survives across polls).

### Tests

`table-run.test.ts` gates the grammar against **every** capture in `fixtures/panes` through each harness's own adapter, the way `prompt-select.test.ts` and `conformance.ts` do. That gate is what found the T-piece false positive; every hand-written case passed while 18 fixtures were being claimed. Result now: claude 50 captures / 1 claim, and that one is a real 200-column table in the scrollback of `claude--trust-prompt.txt`; codex 16 / 0; omp 22 / 0; grok 25 / 0; agy 8 / 0.

`ansi-output.test.tsx` adds the arrangements where the hoisted newline decides — a run that starts the mirror, one that ends it with no trailing newline, two runs, two adjacent runs — with find and link offsets asserted by position rather than by text, plus a guard that the scroller survives a poll.

`cd web && bunx vitest run src/lib/table-run.test.ts src/components/ansi-output.test.tsx` gives 47 passing; `bunx tsc --noEmit` is clean.

### Fork PR housekeeping

Version files and CHANGELOG deliberately untouched; `scripts/check-version.sh` prints its ok line at 1.2.0. Proposed CHANGELOG entry, in your hands:

> **Added** — A table in the mirror pans in its own scroller while the prose around it keeps wrapping, so Wrap no longer has to be turned off to read one (#5).

The Wrap hint changed in all six locales, because it told the operator to turn Wrap off for a table. The stale copy of that rationale in `use-display-prefs.ts` is corrected too.

### An ADR, if you want one

I have left `.adr/` alone — a number claimed in a fork collides with your next one the same way a guessed version bump collides with the release commit, and I cannot accept my own. But I think one is warranted, as an **amendment in scope to 0008**, not a supersede: 0008 stays entirely correct, Collie still runs no emulator, but its Decision says a column-faithful grid "reopens the pan-vs-wrap question 0.21.0 settled toward wrap", and the code now says "except for a detected table". Two of the roads this closes have no line to attach a comment to, because their subject is code that does not exist: a user-facing toggle (the point is that the operator is not asked) and detect-by-width (there is no width test). Draft:

- **Context** — #5 asked for a per-table scroller and got a global no-wrap default; #53 argued the other way and flipped that default in 0.21.0; part 2 of #53 was closed by measurement; 0008 refused the emulator.
- **Decision** — Wrap stays the default. A detected table gets one horizontal scroller, never one per row. Detection is by anchor grammar, never by width. The recoverable range is phone-width to pane-width only, because `pane.read` returns a rendered grid.
- **Consequences** — A false positive pans readable prose, so the anchor alphabet is narrow and the corpus gates it. A table the terminal already broke is split rather than repaired.

If you would rather not, the fallback is to cut the CLAUDE.md bullet to the rule plus a pointer and leave the argument in `table-run.ts`'s header, which is where it already lives.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -317,6 +317,12 @@ lint guard or the pack-wire guard.
   `terminal session observe`/`control`: a stale mirror is a transport problem, cursor position is an
   upstream ask, and `control` resizes the *shared* PTY
   ([ADR 0008](./.adr/0008-collie-does-not-run-a-terminal-emulator.md)).
+- **A table pans; the mirror around it keeps wrapping** — `lib/table-run.ts` groups a table's rows
+  into a single scroller inside the wrapping `<pre>` (`ansi-output.tsx`). One scroller per table,
+  never one per row. Each grammar anchors on a row nothing else prints — a markdown delimiter row,
+  a `+---+` rule, a frame row carrying a **cross** — and then grows by agreement, so a menu, a
+  chrome box or a rule beside a table is never claimed. `table-run.test.ts` gates it against every
+  capture in `fixtures/panes`; the argument sits in `table-run.ts`'s header.
 - **Never use a `dark:` variant inside the mirror `<pre>`** — it tracks the root theme, which is
   backwards in a surface that renders dark under every theme and inverts in light
   ([ADR 0002](./.adr/0002-invert-the-light-terminal-mirror.md)). Fails silently;
--- a/web/src/components/ansi-output.test.tsx
+++ b/web/src/components/ansi-output.test.tsx
@@ -75,7 +75,8 @@ describe("mirror line wrapping", () => {
     const text = `ordinary prose\n${ESC}[41m${border.slice(0, 12)}${ESC}[44m${border.slice(12)}${ESC}[0m\nsee https://herdr.dev/docs\n`;
     const { container } = render(<AnsiOutput text={text} query="───" />);
     const pre = container.querySelector("pre")!;
-    const clipped = pre.querySelector("span.inline-block")!;
+    // `span.overflow-hidden`, not `span.inline-block`: the table-run scroller is an inline-block too.
+    const clipped = pre.querySelector("span.overflow-hidden")!;
 
     expect(clipped.className).toContain("max-w-full");
     expect(clipped.className).toContain("overflow-hidden");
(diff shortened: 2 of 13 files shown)


AltanS:
Thanks @enieuwy. Merged locally as 8d079ff.

This landed on a main that had gained FRAME_ROW (#156): a row whose first
and last non-space glyph are frame edges gets clipped to one visual line.
Every row of a box-drawn table is exactly that shape, so both rules fired
at once and the table refused to pan. Follow-up c81964b sets the
precedence: a detected table run wins on the rows it owns, frame rows
included. Outside a run nothing changes, so a lone framed row or a menu
border keeps its clip. renderLine now takes one `inRun` flag to carry the
order; the AST is untouched. rule-glyphs.ts keeps both alphabets, the
FRAME_EDGE class from main and your three box-table classes, with the
overlap between the two questions written down in its header.

If you want to write the ADR 0008 amendment for that precedence rule, the
floor is yours. Happy to review it.

PR #160, opened 2026-09-03
fix(omp): recognize rule composer shape

taiansu:
## Summary

- recognize OMP 18.1.10's configurable `composer.shape: rule` layout alongside the existing OMP 17 boxed composer and OMP 18.1.2 clipped-box rendering
- recover empty, single-line, and wrapped drafts; reuse the existing styled-run heuristic so inline completion suggestions do not become draft evidence
- bind destructive clears to the rule composer's complete visible prompt region and re-surface its standalone status row
- add three byte-faithful OMP 18.1.10 sandbox fixtures, explicit tail/cap/rejection coverage, versioned corpus documentation, and a changelog entry

Scope follows the maintainer guidance in #161: this PR handles the `rule` shape only. The scanner remains inside `harness/omp/`; it does not share a locator with the Claude harness and adds no interactive OMP dialog or keystroke recipe.

## Root cause

#150 fixed a clipped OMP 18.1.2 box rendering whose bottom prompt row appears as an open-ended `╰─ <draft>` form. The `rule` renderer is a distinct shape with no `╰─` anchor:

```text
────────────────────────
❯ <draft>

<status>
```

The current OMP adapter therefore returned `composerReady: false` before typing. A forced retry could type the text, but `extractInputDraft` still returned `null`, so the guard correctly withheld Enter and reported that the message did not reach the input box.

Renderer source: https://github.com/can1357/oh-my-pi/blob/v18.1.10/packages/tui/src/components/composer/rule.ts

Related failure class: #149 / #150. This PR covers only the separate `rule` shape.

## Safety

The scanner does not search for a loose `❯` row. It requires the complete renderer choreography at the buffer tail: a top rule directly adjacent to the prompt, no more than 100 two-space continuation rows, exactly one blank gap, and one standalone status row as the final non-blank row.

Tests pin both sides of the hard cap (100 accepted, 101 rejected), tail anchoring, every captured OMP menu/select modal, and every Claude, Codex, and Grok fixture as rejection cohorts. The prompt binding contains every visible prompt row, including the visible ghost suffix. Draft verification removes that suffix only from the extracted input value; Enter remains gated on read-back evidence.

The original `## omp corpus` section remains the provenance for the OMP 17.2.12 captures. The new captures have their own dated OMP 18.1.10 section.

## Verification

- `cd web && bun run test` — 166 files passed; 4,956 tests passed; 30 existing todos
- `cd web && bun run typecheck`
- `bun run lint`
- `cd web && bun run build`
- three fresh `pane.read format:ansi` captures from an OMP 18.1.10 sandbox using `composer.shape: rule`: idle, single draft, and wrapped draft with a styled inline suggestion
- fixture hygiene: CRLF preserved, no trailing newline, and no account, host, home-directory, session, UUID, or credential-shaped data

Live guarded-reply verification used this checkout's real Collie UI against an OMP 18.1.10 sandbox:

1. `COLLIE_RULE_18110_STALE` was stranded in the terminal composer.
2. The destructive clear was sent through `/keys` with `expected_prompt: "❯ COLLIE_RULE_18110_STALE"`.
3. Collie typed `COLLIE_RULE_18110_LIVE_ACK_FINAL` with `submit:false`, read the pane back, then sent the empty `submit:true` request.
4. OMP rendered the exact marker without the stale prefix.
5. With `/model` open, Collie retained `COLLIE_RULE_18110_MODAL_GUARD`, offered the explicit `Type anyway?` override, sent no `/reply` or `/keys` write, and left the modal unchanged.


AltanS:
Thanks Tai An, this lands. Your two commits are on main from 72fbfece with your authorship, plus one of mine that names the Herdr version in the fixture header and adds the changelog credit. It ships in 1.6.0.

Issue #169, opened 2026-09-06
Herdr-plugin installs need Bun to install and to update, though the release already ships a built binary


mikebenner:
## Summary

A `herdr plugin install` Collie needs Bun **to install** and needs it again **on every update**, even though the release for that platform already publishes a compiled `bin/collie`. On a host where the bridge's service environment cannot see Bun, the update preflight is permanently red and the phone tells the operator to install Bun.

The preflight is not lying. On the phone-initiated path the update genuinely would stop after the fetch. But the requirement itself looks avoidable: the artifact Collie would compile is sitting on the release page, built and checksummed.

Observed on 1.5.2 (first seen on 1.5.1; the code involved is unchanged between them), Linux x64, Herdr 0.8.2.

## What happens

`GET /api/update/check` on a healthy managed install:

```json
"installKind": "detached-checkout",
{"id":"bun","verdict":"red",
 "reason":"bun is not installed, and this install rebuilds from source — the update would stop after the fetch",
 "remedy":"install Bun from https://bun.sh, then re-run this check"}
```

Every other check is green. The service's own environment is:

```
PATH=/usr/local/bin:/usr/bin
```

Bun is installed on this machine, but only through a version manager's shim directory. `bridge/tools.ts`'s `fallbackDirs` searches `~/.local/bin`, `~/.bun/bin`, `~/.cargo/bin`, `/usr/local/bin`, `/usr/bin`, `/bin`, the sbin dirs and `/opt/homebrew/bin`. None of those holds Bun here, so the check is correct on this path. Invoking the same update from a Herdr action in a terminal *does* work, because Herdr's environment has the shim directory on PATH. One install, two answers, depending on which surface you touch.

## Root cause

`classifyInstall` (`cli/install-kind.ts`) decides the kind from the shape on disk, and any git checkout becomes `linked-clone` or `detached-checkout`. `buildsFromSource` (`cli/update-check.ts`) then treats both as compiling, which gates `bunCheck`.

`herdr plugin install` creates its checkout with `git init` + shallow fetch + detach, so **every** turnkey Herdr install of Collie carries a `.git` and is classified as a source checkout. It also carries a finished `bin/collie`, built once by the `[[build]]` step. It is a git checkout and a delivered artifact at the same time, and only the first fact is being read.

## This also blocks a clean install

On a machine with no Bun at all, `herdr plugin install AltanS/collie` cannot complete. `scripts/collie-ctl.sh` refuses at the bootstrap:

```
[...]

PR #171, opened 2026-09-06
feat(cli): recognise a root-owned install, and defer its updates

mikebenner:
Closes nothing on its own — this is the **first** of the two proposals in #169.

> **Note.** This does **not** silence the Bun warning on a `herdr plugin install`. That install is a
> git checkout that genuinely does rebuild, so `buildsFromSource` stays true for it and the check
> stays honest. #169's second proposal — the plugin channel taking the published payload instead of
> compiling — is what retires it. Stated up front because that warning is the symptom the issue
> opens with, and a reader should not have to reach the bottom of this description to learn it is
> still there.

## What this is

Collie had three install shapes and every one of them updated itself. This adds the fourth that
exists on real machines and had no name: Collie sitting in a directory owned by root, put there by a
distro package or unpacked as root by hand.

Before this, that shape fell out of `classifyInstall` as `{kind: "unknown", why: "loose-binary"}`,
so `collie update` refused with *cannot tell how this Collie was installed* and `doctor` warned
about an unrecognisable tree. Both were the right instinct expressed as a diagnosis failure — an
operator whose install was working perfectly was told it was broken.

## The signal is ownership, and it took two tries

`stat(root).uid === 0`. Ownership is a shape on disk in the same sense as a `.git` directory or a
`versions/` parent, so `cli/install-kind.ts` keeps the structural rule its own header sets: nothing
writes a marker and nothing reads one as the primary signal.

**The first cut used `access(root, W_OK)` and it was wrong**, which is worth stating plainly because
it reads well and someone will propose it again. `access(2)` is always true for uid 0, so the same
tree classified two different ways depending on how the command was typed: `collie update` saw a
system install, `sudo collie update` saw `loose-binary` and printed the exact sentence this kind
exists to delete. A bridge running as root — an ordinary container — never saw the kind at all.
Ownership is a fact about the tree rather than about who asked.

Two orderings carry weight, and both are pinned by tests:

- **Ownership is asked LAST**, only where nothing else claimed the tree. A root-owned clone is still
  a clone whatever `chown` last did to it.
- **The marker still outranks it.** A root with no `herdr-plugin.toml` stays `no-marker` however it
  is owned — that is a directory that is not a Collie. A null uid is likewise not system-owned:
  nothing could be read, so nothing may be asserted.

## It states the ownership and never the provenance

This is the second correction and the one with teeth. Root ownership is observable; *that a package
manager put it there* is an inference, and a tarball unpacked as root is indistinguishable.

The first draft asserted "was installed by a package manager, which owns its updates", which sends
whoever unpacked it themselves after a package that does not exist, while the real remedy — reinstall
the same way, or take ownership — goes unmentioned. Every message now names the fact, offers both
exits, and names no manager as *the* command. One exported constant, `SYSTEM_OWNED_REMEDY`, replaces
what had become five separate spellings.

Note the consequence deliberately kept: running as root, Collie *could* write the tree. It still
refuses, because overwriting a package manager's files leaves its database lying about what is
installed.

## What changes behaviourally

| surface | before | after |
| --- | --- | --- |
| `collie update` | *cannot tell how this Collie was installed* | states the ownership, offers both exits, exits 1 |
| `collie update --rollback` | a lecture about `versions/` and `git checkout` | the same boundary |
| `collie doctor` | `warn: install … cannot tell` | `✓ install  … is owned by root …` |
| doctor `versions` | "the next `collie update` stages one" | skipped — it stages nothing here |
| doctor `update-source` | `github.com/AltanS/collie` | names the ownership, not a repo it never fetches |
| preflight | `doctor, disk, upstream, service` | `doctor, install, upstream, service` |
| upstream remedy | `collie update --major` | the shared remedy — **and only when it named `collie update`** |
| `POST /api/update` | accepted, then spawned an update that refuses | refused with `update.system_owned` |
| footer banner (major) | `collie update --major` | the line and link, no command |
| Updates card | an enabled button whose tap fails | disabled, with who does update it |
| a packaged pack lead | could not level its peers at all | `retry-pack` — peers are a different act |
| `pack update` walk | pushed a git bundle at a root-owned peer | skips it with its own sentence |

The preflight list is shorter and each omission is a claim: no bun because nothing compiles, no tree
because there is no working tree, no disk floor because that floor exists for a staged payload.

**The remedy rewrite is narrow on purpose.** An earlier cut replaced *every* upstream remedy, so an
offline host was told to run its package manager instead of "check this machine's network" — and
because that check is red, the clobbered sentence became the blocking reason the phone displayed.
Only a remedy that actually names `collie update` is replaced now, and there are tests for the
network and rate-limit cases.

## The server gate, not just the button

A system-owned install's preflight is **green** — nothing is wrong with it — so `firstRed` finds
nothing and every other gate on `POST /api/update` passes the start through. The route says of
itself that the client's disabled button is a courtesy and the server is the real gate, and it was
right: a cached bundle, a second tab or a plain POST would mint a run id and spawn a `collie update`
whose only possible outcome is its own refusal, recorded as a failed run on a healthy machine.

The refusal sits **below** the peers-only branch, and a test asserts that placement: a lead that
cannot move itself can still level its peers.

## The pack half

A peer's install kind is **not** a wire field and this does not add one — that would be a protocol
decision under ADR 0025. What already crosses is the member's whole preflight, and `instanceChecks`
emits a check with id `install` on exactly one branch. `PreflightCheck.id` is documented as the
stable identifier the pack flow branches on, so that is what the member walk reads. A system-owned
peer is skipped before a runner is opened, carrying its own sentence, and the run continues for the
members it can level — the `ops-record` shape exactly.

## ADR 0006 is untouched

Its subject is the Herdr-managed checkout, which still advances in place, is still never re-linked,
and still has reinstall as its floor. A root-owned tree is a kind beside it. The re-link prohibition
does not reach it either: the concern was losing `herdr plugin install` as the only remaining
refresh, and a package manager is a stronger refresh than the one it was protecting.

[ADR 0035](.adr/0035-a-system-owned-install-is-not-ours-to-update.md) carries the argument, including
the writability approach recorded as tried and rejected. **The number is yours to reassign.**
`CLAUDE.md`'s update-notice rule was wrong for this kind — it told contributors a non-Herdr install
gets `collie update` — and now names all three.

## A bug this exposed

**The release tarball does not contain `scripts/`.** Verified against v1.5.2: 58 files, none of them
the shim. Every `[[actions]]` command in the shipped `herdr-plugin.toml` is frozen as
`bash scripts/collie-ctl.sh <verb>`, so a package built from that payload installs Herdr action
buttons that cannot resolve. Invisible until now because the payload's only consumer was
`install.sh`, which never registers a plugin. `release.yml` now copies the one shim, and
`docs/install.md` carries a caution for anyone packaging 1.5.2 before the next release.

## The package

`packaging/aur/PKGBUILD` builds `collie-bin` from the published tarball: no Bun, no git, no
compilation, sums from the release's own `.sha256` assets. Herdr already ships this way —
`pacman -Si herdr` names the `omarchy` repository.

`/usr/bin/collie` is a **symlink** into `/usr/lib/collie/bin/collie`, and that is load-bearing.
`bridge/root.ts` resolves the plugin root as `dirname(dirname(execPath))` over a realpath-resolved
`execPath`, so through the symlink the root is `/usr/lib/collie`, where the manifest lives. Installed
straight into `/usr/bin` the root resolves to `/usr` and the bridge finds neither `web/dist` nor the
manifest. Measured: through the symlink the binary reports `1.5.2+cea2035`; alone in a bare `bin/` it
reports `unknown`.

The maintainer line is left unfilled on purpose — nothing in this repo records an address for
whoever would hold that role.

## Verification

```
bun run typecheck            # clean
cd web && bun run typecheck  # clean
bun run lint                 # clean
bun run test                 # 3996 pass, 0 fail
bash scripts/check-version.sh # ✓ 1.5.2 consistent
```

End to end against a real root-owned tree, service-style PATH, binary built from this branch:

```
$ env -i PATH=/usr/bin:/bin COLLIE_PLUGIN_ROOT=$R $R/bin/collie update --check --local --json
checks: ['doctor', 'install', 'upstream', 'service']      # no bun check

$ env -i PATH=/usr/bin:/bin COLLIE_PLUGIN_ROOT=$R $R/bin/collie update
error: … is owned by root, so `collie update` will not replace its files.
```

A real detached shallow checkout on the same binary still reports
`['doctor','disk','bun','tree','upstream','service']` with the bun check red — the unchanged plugin
path above.

One local note, not from this diff: `scripts/collie-cli.test.sh` fails on my machine at the
`collie build` assertion because Bun lives only in a version manager's directory, which
`bridge/tools.ts`'s `fallbackDirs` does not search. Proven environmental — with Bun reachable the
assertion passes.

## How this got here

Three review passes, each catching what the last one missed, and this description already reflects
the corrected state rather than the history:

1. Widening the `installKind` union satisfied `tsc` while leaving both phone surfaces naming the
   command the branch makes fail.
2. The writability predicate, the remedy clobbering, the missing server gate, and the pack gaps —
   the design-level defects this description leads with.
3. **A real gate bypass, a platform bug, and a UX ordering bug**, all found after round 2's fixes
   were already pushed:
   - `bridge/pack/follow.ts`'s pack-follow mechanism never checked install kind, only
     `firstRed(preflight)` — and a system-owned preflight is green by design. A system-owned peer
     following its lead sailed through every guard and spawned a `collie update` that could only
     refuse, once an hour, forever. Fixed with a guard ahead of every other check, sourced from the
     same probe the bridge already runs at boot.
   - `realFiles.ownerUid` reported `0` on win32 — Windows has no POSIX uid, and Node/Bun's `stat()`
     answers a constant `0` there regardless of who owns the file. Colliding that with root would
     have misclassified an ordinary win32 install as system-owned. It now answers `null` on win32,
     the same as a failed `stat`.
   - The update card checked the package-manager sentence before a genuinely red check, so an actual
     fault (a broken `doctor` or `service` check) hid behind "your package manager updates this" —
     true, and useless for finding the real problem. Reversed.
   - Two doc claims restated exactly what ADR 0035's own Context section names as the rejected first
     draft — a stale example error message, and the writability framing the record spent a section
     correcting. Both fixed to match the code.
   - `SYSTEM_OWNED_CHECK_ID` is now one exported constant instead of two files agreeing on the string
     `"install"` by convention, and `cli/update.ts`'s own refusal — the most user-visible of the
     three spellings this shape had accumulated — now composes from the same shared helpers doctor
     and the preflight already used.

The intermediate commits are on the branch rather than squashed, so the reasoning at each stage is
readable.

## Fork PR

Per `CLAUDE.md`, the three version files and `CHANGELOG.md` are untouched. If you want a CHANGELOG
line in my words:

> `collie update`, `doctor` and the update API recognise an install owned by root and defer to
> whatever installed it, instead of refusing as unrecognisable (#169)


AltanS:
Thanks Mike, this is great work, and it lands. I have cherry-picked all 16 of your commits with your authorship onto a branch, and I am adapting a few things on top so it fits the packaging milestone I have been drafting from your issue 169: the kind is called `packaged`, the detection also treats a read-only root or a root outside `$HOME` as packaged so Homebrew and the Nix store are covered, and the pack lead reads a peer's install kind from a field instead of a check id. Your release tarball fix and the Windows guard go in as they are. I will close this PR when the branch merges and credit you in the changelog and the ADR.

AltanS:
Merged into main as 422e1c241bc7818d721f62b125ea1aa296f4005c with your commits. Thanks again, Mike.

PR #174, opened 2026-09-06
fix: check the systemd user bus is reachable before handing update to systemd-run

chernesk:
## Summary

- `handOff()` (the detached-runner handoff `collie update` uses) decided whether to use
  `systemd-run` by checking only `which systemd-run` — whether the binary exists on `PATH`.
- In a container that ships the `systemd` package but never runs a user manager or session bus
  (no `$XDG_RUNTIME_DIR`/`$DBUS_SESSION_BUS_ADDRESS`), that check says yes, the update stages,
  hands off to `systemd-run --user --collect`, which immediately fails with `Failed to connect
  to user scope bus via local transport`, and the run is left stuck in `staging` forever — every
  retry hits the exact same doomed handoff.
- `supervisionTier()` in `cli/lifecycle.ts` already asks the right question
  (`systemctl --user show-environment` succeeding, not just binary presence) to decide the
  `start`/`stop`/`restart` supervision tier. This PR extracts that probe into an exported
  `systemdUserReachable(exec)` and has `handOff` require it too, so a bus-less container falls
  back to `setsid` exactly as it already does when `systemd-run` is absent outright.

## Suggested CHANGELOG line

- `collie update`'s handoff to `systemd-run` now checks the user bus is reachable, not just that the binary exists, so a container with systemd installed but not running no longer wedges every update behind a doomed handoff.

## Test plan

- [x] `bun run lint` — clean
- [x] `bun run typecheck` (root) and `cd web && bun run typecheck` — clean
- [x] `bun run test` (root) — 2700+ tests pass, including the shell-driven `collie CLI lifecycle`
      suite
- [x] Added a regression test in `cli/update.test.ts` reproducing the exact shape (systemd-run
      binary present, `systemctl --user show-environment` failing) — confirmed it fails against
      the old code (falls back to `systemd-run` anyway) and passes with the fix (falls back to
      `setsid`)

### DIFF
--- a/cli/lifecycle.ts
+++ b/cli/lifecycle.ts
@@ -80,10 +80,23 @@ export type Tier = "systemd" | "launchd" | "unsupervised";
 const TIERS: readonly Tier[] = ["systemd", "launchd", "unsupervised"];
 
 /**
- * Which supervisor runs the bridge. `systemctl --user show-environment` succeeding — not merely
- * `systemctl` existing — is the gate, because a container or a machine with no user instance has
- * the binary and no bus (the pre-shim collie-ctl.sh). launchd is gated on Darwin too: the
- * `gui/<uid>` domain is Darwin-only.
+ * Whether the systemd user manager is actually reachable — `systemctl --user show-environment`
+ * succeeding, not merely `systemctl`/`systemd-run` existing on disk. A container commonly ships
+ * the systemd package (so the binaries are on `PATH`) without ever running a user instance or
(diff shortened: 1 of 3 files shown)


AltanS:
Merged into main as bcda1b4, with the changelog line in cccb71a. Thank you for the clean fix and the regression test, @chernesk. It will ship with the next release.

chernesk:
Thank you for being open to the PR.

Sep 6, 2026 4:05:29 PM Altan S ***@***.***>:

PR #176, opened 2026-09-06
fix: recognize Codex inline queue status

stekman08:
## Summary

Codex can render its queue hint and context percentage on the same raw terminal row while a turn is active. Collie's composer locator previously recognized the older dot-separated/context status-row shapes, so this visible composer was treated as absent and the guarded reply path returned `Message didn't reach the input box` before typing or submitting.

The root cause is the raw pane layout, not the phone's visual line wrapping: the captured ANSI buffer contains one footer row with both `to queue message` and `93% context left`. The parser must recognize that generic row shape before applying the existing prompt, verification, and guarded-submit flow.

## Changes

- Recognize the generic inline queue/context Codex footer row in the Codex marker parser.
- Add a small anonymized ANSI fixture preserving the observed SGR styling and one-row layout.
- Add focused parser and `sendGuardedReply` regressions proving type -> verify -> submit ordering.
- Keep dialog detection, text verification, prompt binding, and guarded submit behavior unchanged.

## Red-first evidence

With the pre-change parser, the new fixture produced `locateComposer returned null`. With this change it locates `{ promptRow: 2, statusRow: 4 }`, and the focused send regression passes with submit still guarded.

## Scope

This PR is limited to the Codex parser, its fixture, and the focused regression tests. Release metadata is intentionally untouched, as required for functional changes submitted from a fork.

## Verification

- Backend and web typechecks: passed.
- Lint, version consistency, and `git diff --check`: passed.
- Focused Codex/reply-action tests: passed.
- Full web Vitest suite: passed.
- Bridge root, mux, journal, pack, CLI, and scripts test groups: passed when run independently.

The local full backend invocation also reaches all assertions in the STT suites but does not terminate under the installed Bun 1.3.14/macOS runtime because two existing response-body disposal tests leave a deliberately pending stream cancellation; the changed files are outside that code. The pushed PR's CI run is the authoritative full-suite check.

### DIFF
--- a/web/src/fixtures/panes/README.md
+++ b/web/src/fixtures/panes/README.md
@@ -36,6 +36,7 @@ and the observed commands were approved with no dialog painted.
 | `codex--fresh-idle.txt` | Welcome banner box, tips, empty `› Ask Codex to do anything` composer, status row | `idle` |
 | `codex--draft.txt` | One-line draft on the `› ` row | `idle` |
 | `codex--draft-wrapped.txt` | Long draft word-wrapped onto a two-space-indented continuation row | `idle` |
+| `codex--queue-context-inline.txt` | Queue hint and context percentage share one raw footer row while the composer remains visible | `working` |
 | `codex--working.txt` | `• Working (3s • esc to interrupt)` above a still-visible composer (Codex queues mid-turn) | `working` |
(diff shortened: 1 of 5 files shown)


AltanS:
Thanks Fredrik, this is exactly the kind of fix that lands, and it shipped in 1.5.5. Your commit went in as a cherry-pick with your authorship intact (0ad4f2f), and the changelog credits you and this PR. Your diagnosis is right: the raw footer row carries both the queue hint and the context metric, and the parser had to know that shape before the guarded path could see the composer. One thing landed on top of it: 1.5.5 also binds the guarded submit to the verified prompt (#177), so your inline-queue fixture test now expects the bound prompt as well, which I wrote out in a follow-up commit. Closing this since the branch is on main through the cherry-pick.

PR #177, opened 2026-09-06
fix: bind guarded submit to verified prompt

stekman08:
## Summary

- Bind guarded submit to the prompt region verified immediately after typing.
- Forward that region as `expected_prompt` so the bridge rejects a stale submit with `prompt_changed` before sending submit keys.
- Preserve the verified draft and existing screen-changed error behavior.

## Tests

- `web/src/lib/reply-action.test.ts`: verifies the binding and the prompt-change race, including that text remains unsent.
- Focused reply-action tests: 42 passed.
- Backend and web typechecks plus lint passed.

This fork PR intentionally leaves version files and `CHANGELOG.md` unchanged, per `CONTRIBUTING.md`.

### DIFF
--- a/web/src/lib/reply-action.test.ts
+++ b/web/src/lib/reply-action.test.ts
@@ -16,16 +16,19 @@ const paneWithDialog = "Do you want to proceed?\n ❯ 1. Yes\n   2. No\n\n Esc t
 
 /** Record every reply POST, and let the fake pane's screen be swapped per test. */
 function harness(screen: () => string) {
-  const calls: Array<{ text: string; submit: boolean }> = [];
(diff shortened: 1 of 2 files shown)


AltanS:
Thanks Fredrik, this lands too, and it is in 1.5.5. Cherry-picked with your authorship kept (e7c1c78) and credited in the changelog. You closed a real gap: the bridge already checked expected_prompt and already answered prompt_changed, and the guarded path simply never sent it, so a dialog that took focus between the typing and the Enter was answered instead of refused. I verified the bridge side live against a matching and a stale prompt before tagging. Closing this since the commit is on main through the cherry-pick.

PR #182, opened 2026-09-07
feat(zen): rotating to landscape can open zen

enieuwy:
Turning the phone sideways enters zen; turning it back leaves. Behind its own toggle, default off.

Landscape is where chrome costs the most, because a narrow tall mirror becomes a short wide one and every chrome row eats terminal lines. Reaching the pane menu to tap zen, in the orientation where a control is hardest to hunt for, is the part this removes.

### Behaviour

- The effect acts on a rotation *flip*, never on a re-render, so tapping the floating way out in landscape stays out instead of instantly re-entering.
- A hand-entered zen is never stolen, because the effect marks its own entries, and only a marked one exits on the way back to portrait.
- Mounting already sideways already opens chrome-free, matching a reload with the phone in hand.
- Every entry runs through `enterZen`, not a bare `setZen`, so the find bar, the sheets and a staged key queue clear exactly as on a manual entry.
- Both bits gate it — zen availability and the landscape sub-toggle. Either one going off mid-zen exits a marked entry rather than stranding you in it.

### The setting

A dependent row inside the existing Zen card rather than a card of its own, since it is the same feature at a finer granularity and is meaningless while zen is unavailable. It stays visible and disabled while zen is off, because hiding it would let a preference flip unseen.

Its text hangs under the card's title rather than under the icon (`pl-12` = the card's `px-4` plus the 32px icon gutter), so one left edge runs down the card. `notify-prefs.tsx` keeps its rows flush at `px-4` instead; those are peers under a shared header, while this is a dependent of a titled row.

### Notes

`use-media-query.ts` reads once for the first render and re-reads on subscribe, so a flip landing between render and effect is not lost. The active zen state stays transient local state in `AgentChat`, reset by the `key={paneId}` remount, so a pane always opens normal.

### Tests

`web/src/lib/zen.test.ts` covers the persisted bit and its default; `agent-chat.test.tsx` covers entry on flip, exit on flip back, a hand-entered zen surviving a rotation, and the mid-zen gate. 13 files, all seven locales. No version bump or CHANGELOG entry, per CLAUDE.md's fork rule.

Suggested CHANGELOG line, if you want one in my words:

```
### Added
- Rotating the phone to landscape can enter zen automatically, behind a Settings toggle
```

I have run this on my own phone for a while, but it is your look and your call — happy to change or drop anything here.


enieuwy:
### Screenshots

Hosted on a branch of my fork rather than committed here, so the diff stays code-only. The sub-row is present and disabled while zen is off, because hiding it would let a preference flip unseen:

| Zen off | Zen on |
|---|---|
|  |  |

The rotation itself is not photographable in a still; `agent-chat.test.tsx` pins it instead — entry on a flip, exit on the flip back, a hand-entered zen surviving a rotation, and the mid-zen gate.

AltanS:
Thanks Ellis, this lands. Your three commits are on main from 36d7b974 with your authorship, plus one of mine: the automatic flag is cleared on a hand exit so a zen you re-opened yourself survives turning back, the query gained a max-height term so desktops and tablets stay out, and the switch defaults off. It ships in 1.6.0.

PR #183, opened 2026-09-07
feat(keys): a dense key layout that gives the mirror back its rows, behind a per-device toggle

enieuwy:
A second, opt-in layout for the key surfaces: compact grids and one always-on key rail. What it buys is mirror, because every chrome row is a row of output you cannot see.

The gain is largest while driving a menu or passing hotkeys, since that work is a run of taps against a screen redrawing under each one — the keys have to stay on the glass, and the roomy layout pays for them by covering the thing you are steering. Nothing here is about screen size: a large phone spends the same rows on chrome and gets the same rows back. The reference is a phone terminal like Termius.

Default off, so an install that never opens Settings renders exactly what it renders today.

### The toggle

One bit, per device (`lib/density.ts`, `localStorage`), deliberately not a scale: the dense choices only read together, since smaller targets buy more keys per row, which is what makes a single rail worth its row. A per-surface dial would allow combinations nobody laid out. Nothing measures the screen, so this is a taste rather than a capability, and two devices paired to one collie are meant to disagree.

**Side pins** — which edge the /Agents and Keys tabs dock to — is that card's dependent row, present only while the layout with pins to place is on.

### What the dense layout draws

- A key rail above the reply box, always on screen, carrying Escape, Tab, ⇧Tab, the four arrows, ^C, and a pad that opens the full tray. It replaces the roomy Controls row outright, and the pad is the only thing costing a tap.
- The full tray holds 69 keys in 8 columns, two lanes, with numpad and symbol rows.
- Under the mirror sits a sessions row: every agent in the space as a scrollable run of titles, the /Agents palette pinned to the configured edge, and a grab handle to the switcher sheet.
- Quick replies fold into the agent palette, so the Quick dock is not a second door.
- Chrome shrunk to match, since a denser key surface is pointless if the frame keeps the room: header floor 60px → 52px, reply box 44px → 36px, the status band and the 30px grab bar stand down (both were duplicated by surfaces this layout already has), and the docks lose their header row, share one 30dvh cap, and draw one boundary instead of two.

Every roomy surface is untouched. The two layouts are separate components where they differ, so the setting cannot half-apply.

### Two fixes worth calling out

The dock site glides as one box: all docks share a `Collapse`, so switching drawers is a single height change rather than an unmount racing a mount. The sessions row snaps out as a dock opens and glides back alongside it closing, so the bottom edge only ever moves one way — a row gliding out beside a dock gliding in reads as a bounce against the live-wrapping mirror. Each opening is keyed, so a reopen mid-exit remounts instead of reconciling onto the held dock and resurrecting a discarded key queue (ADR 0005 surviving the glide). Roomy keeps its existing motion exactly.

The switcher handle is now the house grab bar. It had been a pill straddling the border, and since a real border cannot survive a `clip-path`, its edge was a doubled drop-shadow silhouette with a pinched hexagon to make the direction legible. It is now the same bar `agent-chat.tsx` and `ui/sheet.tsx` already use for "this surface opens", sitting wholly on chrome tight under the rule. No pixels added: it is absolute inside the 10px lane the row already reserves.

### Verification

Typecheck clean; the suite passes on this branch, merged onto `main`, and with the zen PR merged alongside. Both layouts are pinned by tests, including the dock contract as one decision (headerless, ruleless, one shared cap), so a partial regression fails. Geometry and motion were measured against a reference build in parallel worktrees — computed styles and bounding boxes in three dock states — rather than eyeballed.

No version bump or CHANGELOG entry, per CLAUDE.md's fork rule. Suggested lines, if you want them in my words:

```
### Added
- Dense key surfaces: a compact key tray, an always-on key rail and a sessions row, behind a
  per-device Settings toggle
- Side pins: which edge the /Agents and Keys tabs dock to

### Changed
- Dense layout trims the chrome around the mirror: header floor, reply box, status band, grab bar
  and dock headers

### Fixed
- The pane switcher handle is the same grab bar the sheets use, under the rule
```

This is a big diff for a look you own, so treat every choice in it as a proposal. Happy to split it, drop parts, or rework anything that does not fit where you want the app to go.


enieuwy:
### Screenshots

Hosted on a branch of my fork rather than committed here, so the diff stays code-only. Same pane, same fixture, same 370×595 crop of the playground's blocked-pane card; taken from this branch's head, not from an earlier revision.

| Roomy (default, unchanged) | Dense |
|---|---|
|  |  |

Dense gives the mirror four more lines in the same box: the tab strip and pane strip are gone, the status band and the 30px grab bar stand down, the header floor and the reply box shrink, and the Controls row is replaced by the key rail — which is more keys on screen than the row it replaced, not fewer.

The tray, open on the same pane. 69 keys in 8 columns, two lanes, and the rail stays under it, so the keys you use most never move:



The setting, and its dependent row. Side pins appears only while the layout that has pins to place is on:

| Off | On |
|---|---|
|  |  |

The shots come from `web/src/playground/`, so any of them can be reproduced without a live bridge.

AltanS:
very intriguing concept! going to think about this a bit. but making the UI more focused and lean is a big goal of mine, so i'll def go through this very thoroughly and take my time. 

again a really great contribution from you @enieuwy thanks!

AltanS:
Thanks Ellis, this is careful work and I want it in, in 1.7.0 rather than 1.6.0, which is a distribution release already gated on Omarchy. Two things first. The rail and tray keys measure 29 to 36 pixels and DESIGN.md sets 44 as the floor for anything tappable, so give them the floor as hit area the way labelled-strip does, and correct the comment in chat-input.tsx that says every dense surface keeps min-h-11. And move ^C next to Escape, since at 390 pixels it sits behind the fade with Enter and Space, and ^C is the reason the rail exists. With those two it lands as is with your authorship.

enieuwy:
Thanks @AltanS! I did take some time iterating, and I'm pretty happy with the outcome. 

Addressed in 6579a88. The rail and tray keep their small key faces with 44×44px hit areas. The tray gaps prevent overlapping targets; below 384px, the grid scrolls sideways instead of shrinking them. ^C now follows Escape, and the chat-input.tsx comment no longer claims every dense surface uses min-h-11.

Browser checks cover all 79 buttons at 320, 375, 390 and 430px, with pins on either side. Taps outside the key faces reach the correct controls. Lint and both typechecks pass. The full web suite reports 5,493 passing tests and the existing transcript date-format failure.

I also merged current main and resolved both conflicts.

AltanS:
Thanks @enieuwy, this was a great inspiration, and it shaped where the pane screen went. I ended up building it another way. The keys and harness shortcuts now sit on one actions belt above the keyboard, and the switcher grip rides on that belt, so the dense layout does not ship as its own setting and I am closing this PR. Your point about giving the mirror back its rows is what started that redesign, and the 44 pixel hit areas you worked out carry into it.

PR #185, opened 2026-09-07
feat(pane): show the newest reply in full when the terminal clipped it

sd2k:
An agent's TUI runs on the alternate screen, which keeps no scrollback ring, so `pane.read` can only ever return the visible viewport. A reply longer than the pane is tall reaches the mirror with its opening already gone, and the only copy that still exists is the agent's own session log — so reading a message you are already looking at meant leaving the pane for the history route.

The pane view now reads the newest turn from the journal and, when that turn *is* the message on screen and its start is missing, renders it in full **in place of** the rows it covers. Everything below the reply — tool calls, a dialog, the cursor — is untouched, so the mirror still reads as the live screen, just starting where the message finished.

### The two decisions worth reviewing

**The gate is an identity check, not a length check** (`web/src/lib/latest-reply.ts`). `locateReply` folds both texts to letters and digits only — which is what makes Markdown source comparable to a hard-wrapped, SGR-coloured render, and is what makes the wrapping irrelevant — then runs two probes:

- The **tail** probe asks whether this journal turn is the message on screen at all. It often is not: the log lags a reply that is still streaming, the operator may have frozen the mirror by scrolling, and a part past `MAX_TEXT_CHARS` arrives clamped so its real ending never does. A missed tail renders **nothing**. Presenting an older message as "the reply you are looking at" is the one failure this must not have.
- The **head** probe then answers the actual question: is its opening still on screen.

**The hiding is render-only.** `locateReply` also returns where the clipped rows end, and `AnsiOutput`'s new `hideLeadingLines` drops them — applied *after* every grammar has run over the whole screen, so no detector, race guard or draft probe ever sees a trimmed buffer. Counting in screen rows is sound because every harness's `stripChrome` slices from the tail only and every pipeline lifts its dialog off the tail, so the leading raw block still starts at row 0; codex's `decorateCodexDisplay` is one-to-one. Collapsing the card gives the raw rows back, which is why the open state lives in `AgentChat` rather than the card — the hiding and the folding have to be one decision. Find restores the whole mirror and stands the card down, since find can only highlight there.

### Cost, and the opt-out

Fetching waits for the mirror to **settle** rather than riding the 1.5 s poll — a streaming reply never settles — so it is roughly one journal read per finished message, the same cost as opening the history route once, which is what it replaces. New `expandClippedReply` display pref, default on, in the ⚙ View dock; off, the pane behaves exactly as it did before and costs no journal read at all.

### Notes

- No version bump and no `CHANGELOG.md` line, per CLAUDE.md's fork rule. Suggested Unreleased line:
  `- A reply whose opening scrolled off the terminal is shown in full from the agent's own log, in place of the rows it covers; ⚙ View → Full latest reply turns it off`
- Five new message keys, added to all seven dictionaries. The six non-English ones are my own work and would benefit from a native reader's eye — `zh-TW` in particular, which I wrote in that file's own Traditional register rather than converting the Simplified entries.
- No ADR: this closes off no option. It respects ADR 0008 — the mirror stays the mirror, and nothing here synthesises scrollback.
- Verified on a real install (Herdr, 1.5.x) against live Claude panes, not only in tests.

### Tests

`locateReply`'s three verdicts over wrapped / emphasised / `⏺`-prefixed screens, the clamped-part case, and `endLine` (including that it reports `-1` on any non-clipped verdict, so a caller that forgets to check hides nothing); `dropLeadingLines`; and seven `AgentChat` cases asserting against the `<pre>`'s own text — rows gone while expanded, back on collapse, back while find is open, nothing on a stale journal read, and no journal read at all when the pref is off or the pane has no log.

5177 web tests and 4028 bridge tests green, both typechecks clean, `scripts/check-version.sh` ✓ at 1.5.5.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -300,6 +300,19 @@ graph TD
   and renders a window that grows upward, which is what lets find-in-history and jump-to-user-turn
   work across turns you haven't scrolled to. Rationale and the measured numbers are commented at the
   top of `web/src/routes/history.tsx`.
+  - **The same source also repairs the pane view in place.** A reply longer than the pane is tall
+    reaches the mirror with its opening already scrolled off, and leaving for the history route to
+    read it is a heavy answer to a small question. So the pane view reads the newest turn from the
+    journal and, when `web/src/lib/latest-reply.ts` says that turn IS the one on screen and its start
+    is missing, renders it in full **in place of** the rows it covers — `locateReply` returns where
+    those rows end and `AnsiOutput`'s `hideLeadingLines` takes them off the top, so the message is
+    never printed twice and the mirror below it (tool calls, a dialog, the cursor) is untouched.
+    Both probes fold the two texts to letters and digits only, which is what makes Markdown source
+    comparable to a hard-wrapped, SGR-coloured render; the tail probe is an identity check, and its
+    failure (a streaming reply, a stale read) means *show nothing*. The hiding is render-only and runs
+    after every grammar, so no detector, guard or draft probe can see it; find restores the whole
+    mirror, since find can only highlight there. One journal read per finished message, off the poll
+    path (`web/src/hooks/use-latest-reply.ts`), and switchable off in ⚙ View.
 - **The browser polls too.** `useRevalidator` → `/api/snapshot` on an adaptive interval. There is no
   WebSocket fan-out to the browser and no push of state; pulling is what makes the two recovery loops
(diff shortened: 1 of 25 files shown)


AltanS:
Thanks Ben, this lands too. Your commit is on main as 46d2fe6a with your authorship, plus one of mine that brings the two new files under the lint rules main gained since you branched and adds the changelog credit. It ships in 1.6.0.

PR #187, opened 2026-09-07
perf(state): skip enrichSessionNames readGrid for unchanged panes

sd2k:
## Summary

- Propagates Herdr's per-pane content `revision` through `MuxPane` → `AgentView`
- `enrichSessionNames` now tracks the last-enriched revision per pane and skips the `readGrid` RPC when the revision hasn't changed
- Drops the per-poll cost from O(claude_panes) socket calls to O(changed_panes)

In a session with many idle/blocked/done agent panes, most panes have an unchanged revision between polls. Before this change, every poll issued a `readGrid` RPC for each Claude pane (one-shot socket connection per pane) to scrape the `/rename` session name. Now unchanged panes are skipped entirely.

### Changes

- **`bridge/mux/types.ts`** — Added optional `revision` to `MuxPane` (Herdr provides it natively; tmux/zellij can only derive it per-read)
- **`bridge/mux/herdr/adapter.ts`** — Propagates `WirePane.revision` through `toMuxPane`
- **`bridge/types.ts`** — Added `revision` to `AgentView` so it flows to the state engine
- **`bridge/state-engine.ts`** — `enrichedAt` map tracks per-pane revision; `enrichSessionNames` skips `readGrid` when unchanged; cleanup on pane removal
- **`bridge/solo-baseline.test.ts`** — Updated exhaustive key checks for the new field

## Test plan

- [x] `bun run typecheck` (root) passes
- [x] `cd web && bun run typecheck` passes
- [x] `bun test bridge/state-engine bridge/solo-baseline` — 92 tests pass
- [ ] Manual: verify session names still appear and update on `/rename` with a live Herdr session

### DIFF
--- a/bridge/mux/herdr/adapter.ts
+++ b/bridge/mux/herdr/adapter.ts
@@ -522,6 +522,7 @@ function toMuxPane(
   // Scrollback depth + viewport = what a `recent` read can yield. Omitted when the server predates
   // `scroll`, so an older Herdr reads as "unknown" rather than "zero".
   if (raw.scroll) pane.readableLines = raw.scroll.max_offset_from_bottom + raw.scroll.viewport_rows;
+  pane.revision = raw.revision;
   return pane;
 }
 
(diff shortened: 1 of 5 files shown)


sd2k:
Superseded by #189, which re-applies this optimization against the current mux-port structure on main. The fork had diverged too far (the `bridge/mux/` abstraction layer was reorganized on main), making a clean rebase impractical — the cherry-pick hit modify/delete conflicts on all three mux files plus content conflicts in the two that survived.

PR #189, opened 2026-09-07
perf(state): skip enrichSessionNames readGrid for unchanged panes

sd2k:
## Summary

Re-applies #187 against the current mux-port structure on main. The original PR was based on an old fork that predated the mux abstraction layer and could not be cleanly rebased.

- Adds optional `revision` to `MuxPane` — Herdr provides it natively; tmux/zellij leave it absent
- The Herdr adapter propagates `WirePane.revision` through `toMuxPane`
- `enrichSessionNames` builds a revision map from the snapshot and tracks the last-enriched revision per pane, skipping the `readGrid` RPC when unchanged
- Drops the per-poll cost from O(claude_panes) socket calls to O(changed_panes)

In a session with many idle/blocked/done agent panes, most panes have an unchanged revision between polls. Before this change, every poll issued a `readGrid` RPC for each Claude pane (one-shot socket connection per pane) to scrape the `/rename` session name. Now unchanged panes are skipped entirely.

Supersedes #187.

## Test plan

- [x] `bun run typecheck` (root) passes
- [x] `cd web && bun run typecheck` passes
- [x] `bun run lint` passes
- [x] `bun test bridge/state-engine bridge/solo-baseline` — 92 tests pass
- [x] Full `bun run test` (bridge + scripts) passes
- [ ] Manual: verify session names still appear and update on `/rename` with a live Herdr session

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ PATH. Details and rollback: [`docs/upgrading.md`](./docs/upgrading.md) → *Upgr
 
 ## [Unreleased]
 
+- `enrichSessionNames` skips the `readGrid` RPC for panes whose content revision hasn't changed since the last enrichment, dropping the per-poll cost from O(claude_panes) socket calls to O(changed_panes) (#187)
+
 ## [1.5.6] - 2026-09-07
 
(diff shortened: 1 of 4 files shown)


AltanS:
Thanks Ben, this lands. Your commit is on main as 198fe204 with your authorship, plus one of mine that adds three tests around the revision guard and the changelog credit. It ships in 1.6.0.

PR #192, opened 2026-09-08
Fix OMP Pi-style composer and multiline paste verification

SocialSolidarityBank:
## Summary
- Recognize OMP's Pi-style editor using its paired rules, indented draft, styled footer and tail anchoring; retain refusal for picker/dialog captures.
- Remove unaccepted inline ghost completions from the Pi draft using the existing `draftGhost` helper. A real Korean capture previously included a suggested suffix that was never typed.
- Keep OMP multiline replies literal by dividing transport pastes into small, lossless parts. Verify each cumulative draft and its tail before continuing, then submit exactly once. Other harnesses retain their single-paste path.
- Treat existing OMP paste chips as opaque in the draft takeover UI; never use a numbered chip as evidence for Enter.

## Reproduction and rationale
With OMP 18.1.13 and `composer.shape: pi`, a 30-line Korean or English paste becomes a compact `📄 #N` chip. The normal literal verification cannot match it, so the message remains unsubmitted until the operator manually presses Enter. Pi-shaped composers were also absent from the existing box/rule recognizers.

The compact chip has neither content nor a length that could safely identify this send. Instead of accepting arbitrary chips, this retains literal verification. The OMP renderer documents marker-sized bracketed pastes (>10 lines or >1000 UTF-16 units); transport parts target 5 lines / 512 units with headroom. Grapheme clusters and exact concatenation are preserved. Artificial path-prefix boundaries are avoided because OMP's paste sanitizer can add a separating space there. An unexpected collapse still fails closed.

## Validation
- Full frontend suite: 192 files, 5418 tests passed, 30 pre-existing TODOs.
- Root lint and both root/web typechecks passed.
- Production builds validated separately on the installed 1.5.3 line and current main.
- Actual Herdr + OMP 18.1.13 integration: a 31-line Korean message was delivered in eight parts, each literal draft verified, followed by one Enter; the agent returned the requested marker.
- Regression coverage: Korean, emoji, CRLF, long lines, opaque chips, model/dialog overlays, transport failure and dropped final pastes including repeated text (no submit). Each multipart proof must differ from the preceding draft, preventing unchanged repeated text from authorizing Enter.
- Independent code review completed; its dropped repeated-text finding was fixed and covered by a regression test.

## Scope and limitations
- Browser-button E2E on the remote host is not claimed: the available test browser is not paired there. The runtime integration used the existing authenticated SSH/Herdr transport; physical iOS IME testing remains outstanding.
- Additional round trips are intentional for OMP large replies. Unknown rendering or paste behavior remains a safe stall rather than blind submission.
- If a genuine append leaves an identical visible window (for example repeated text), verification conservatively stalls instead of submitting.
- Existing local PWA lifecycle and other composer-style experiments are not included. No version, changelog or lockfile changes.


AltanS:
Thanks SeongQ, this lands. The guard keeps its rule, literal verification or no Enter, and your change makes that rule hold on the Pi composer too: ghost completions stripped from the draft, long replies sent in verified parts, and the paste chip treated as opaque. Your commits go in with your authorship kept, the changelog credits you, and 1.8.0 ships it. I will add the changelog line on top rather than ask you for it.

Issue #197, opened 2026-09-09
Add Hermes SessionDB transcript adapter for pane history


jamesjfoong:
## Problem

Collie history paging does not work for Hermes agents running under Herdr. Hermes uses an alternate-screen TUI, so Herdr reports no terminal scrollback (`max_offset_from_bottom: 0`, `readableLines: 50`). Collie 1.6.0 has journal adapters for Claude, Codex, Grok, OpenCode, and pi, but not Hermes.

## Proposed change

- Add a read-only Hermes journal adapter for `~/.hermes/state.db`.
- Use Herdr `agent_session.value` as the exact Hermes session ID; never guess the newest session.
- Read `sessions` and `messages`, preserve user/assistant/reasoning/tool rows, and follow parent-session compression lineage.
- Keep existing transcript byte caps, pagination, cache invalidation, and path containment rules.
- Add `COLLIE_HERMES_ROOT` override and adapter/parser/security tests.

## Environment

- Collie: 1.6.0 (`2e3df8a`)
- Herdr: 0.8.2
- Hermes: local SessionDB at `~/.hermes/state.db`
- Multiplexer: Herdr

## Evidence

The affected pane is detected as `agent: hermes`; current pane history is limited to 50 rows. The Hermes database contains the matching TUI session and hundreds of persisted message rows. A local prototype resolves the exact session ID and parses the transcript successfully.

Would maintainers accept a PR for this, or prefer a Herdr integration/API change first?

AltanS:
very happy to accept a PR for this

jamesjfoong:
Self-review complete against current PR head `55cf0b0aad5c73438a65c88b388da083f9ef1a78` and `AltanS/collie:main`.

**Review result:** no confirmed correctness, security, lifecycle, or architecture blocker found.

Verified:

- Hermes session ID is validated before use; no newest-session guessing.
- Hermes DB opens with `readonly: true`.
- `state.db` stays confined to configured Hermes roots through `containedRealpath`.
- SQL uses bound session parameters.
- Parent-session lineage is bounded to 32 levels.
- Message parsing rejects malformed JSON and invalid row shapes.
- Content/reasoning/tool summaries use existing size and ANSI-stripping helpers.
- Registry, config, frontend allowlist, and server fixtures remain aligned.
- `bun run typecheck`: passed.
- `bun run lint`: passed.
- `bun run build`: passed.
- Focused Hermes/config/registry tests: `62 pass`, `0 fail`.
[...]

PR #199, opened 2026-09-09
feat: add Hermes transcript history support

jamesjfoong:
Hermes panes use an alternate-screen TUI, so terminal scrollback cannot provide older history. This adds read-only Hermes SessionDB transcript support keyed by the exact Herdr `agent_session` value.

**Base branch** — see [CONTRIBUTING.md](https://github.com/AltanS/collie/blob/main/CONTRIBUTING.md#base-branch):

- [x] Opened against `main` — there is one line of development, and `main` is it

### Changes

- Add Hermes SQLite adapter for `~/.hermes/state.db`.
- Resolve only validated Hermes session IDs; never guess from newest session.
- Follow `parent_session_id` lineage for compressed/continued sessions.
- Render user, assistant, reasoning, tool-call, and tool-result rows.
- Open Hermes DB read-only and enforce configured-root containment.
- Add `COLLIE_HERMES_ROOT` override.
- Register Hermes in bridge and frontend journal registries.
- Add parser, registry, config, and security regression tests.

### Related issue

- https://github.com/AltanS/collie/issues/197

### Checks

- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run build`
- [x] Focused Hermes/config/registry tests: `62 pass`, `0 fail`
- [ ] `bun test ./bridge ./cli ./scripts` — timed out in this environment
- [ ] `cd web && bun run test` — existing dependency/runtime failures: `React.act is not a function`

Live validation was performed locally with a real Hermes `state.db` and an exact Herdr session ID; no Hermes writes were made.

### DIFF
--- a/packaging/aur/.SRCINFO
+++ b/packaging/aur/.SRCINFO
@@ -1,6 +1,6 @@
 pkgbase = collie-bin
 	pkgdesc = Phone web UI for the AI agents running in your terminal, served over Tailscale (official binary release)
-	pkgver = 1.7.0
+	pkgver = 1.8.0
 	pkgrel = 1
 	url = https://github.com/AltanS/collie
 	install = collie-bin.install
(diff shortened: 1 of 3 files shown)


AltanS:
Thanks James, this is a clean adapter and it lands. Your commits go in as they are with your authorship kept, and 1.8.0 credits you in the changelog for Hermes history on the phone. Two small things I will do on top rather than ask you for: the registry test title says five harnesses and now lists six, and the changelog gets its line. Issue 197 closes when this merges.

PR #201, opened 2026-09-10
fix(codex): recognize wrapped question cards

alvinycheung:
Codex questions lose their buttons in Collie when the question or description wraps, or `esc to interrupt` gets its own row. I saw this with Codex 0.153.4 and Collie 1.8.0, with Raw terminal off.

This fixes those three shapes in the question parser. It joins wrapped question and description text and keeps the original rows in the tap signature. The digit keys stay the same. Notes mode, incomplete options, and output below the footer still fall back to the terminal.

The tests modify the existing public fruit capture. All four wrapping cases fail against the old parser and pass with this change. These are synthetic variants of that capture, not new live captures.

Checks:

- Both typechecks, full lint, and version consistency pass.
- Web suite: 5,540 pass, 30 todo, including adapter conformance.
- CLI/script tests: 1,486 pass.
- Bridge tests excluding the crew harness: 3,001 pass, one failure in unchanged `bridge/journal/pi.test.ts`. It compares `/var/...` with `/private/var/...` on macOS.
- The full backend run stalled in the crew TLS tests on Bun 1.3.10. I stopped it and ran the other groups separately.

Leaving this as a draft for maintainer live-verification on a wrapped card. I did not send keys to the work session used for the read-only diagnosis. No version or CHANGELOG files changed.

### DIFF
--- a/web/src/lib/harness/codex/ASK_NOTES.md
+++ b/web/src/lib/harness/codex/ASK_NOTES.md
@@ -35,3 +35,18 @@ required: `Question X/Y (N unanswered)` header, a non-empty question line, conse
 pointer rows, and the notes footer. The NOTES-FOCUSED state refuses to raw (footer
 `tab or esc to clear notes`, or a `› Add notes` row): a digit there would type into the box.
 Typing notes from the phone is deliberately not offered — it has no probed recipe.
+
+## Wrapped cards (2026-09-10, Codex 0.153.4)
+
+A read-only inspection found cards with wrapped questions and descriptions. One footer put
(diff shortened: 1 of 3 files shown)


AltanS:
Thanks @alvinycheung, this is careful work, and it lands. I keep your commit as it is. Your joins only happen between the fixed header and footer of the card, and the guard still compares the whole card before it sends a digit. So a wrapped list outside a card cannot be taken for a question. The tests are layout variants of the fruit capture. A capture of a real wrapped card is still the missing piece, and I will check one live before the release. You get credit in the changelog, and GitHub closes the PR when main goes out.

PR #212, opened 2026-09-12
fix(web): link a URL the terminal wrapped as the one URL it was

thelinuxlich:
## The problem

A URL longer than the pane arrives cut at the column edge, because the mirror renders the grid. `web/src/lib/links.ts` scans one line at a time and deliberately does not stitch hard-wrapped fragments, so only the first fragment becomes an anchor — and that anchor's href is truncated mid-URL. The rest of the URL is inert text.

`gcloud auth login --no-launch-browser` is the everyday case: its consent URL is ~300 characters on a 120-column pane, so the one link the operator needs to open is the one link that cannot be opened.

Measured against a live herdr 0.9.0 server (a shell pane, 120 columns, that URL printed into it), same browser, same pane, same request:

| | anchors on the URL | href |
|---|---|---|
| installed 1.8.1 | 2 | 103 and 120 chars — both truncated |
| this branch | 6 | one distinct href, 288 chars — the whole URL |

## The fix

Two halves, because only one of them can know the whole URL.

**Bridge.** herdr can hand over the same rows with soft wraps undone (`recent_unwrapped`). When the grid it already has shows a URL touching a line end (`hasSplitUrl`, so every other pane pays nothing), the bridge asks for that read, strips the SGR (`stripSgr`), and sends it as `PaneReadResponse.logicalText`. One extra read, on the panes that show a split, with the same `lines` and — deliberately — the same `ansi` format as the mirror read: the observation that keeps that read from harvesting is the one the mirror already rests on, and there is nothing else to lean on here.

**Client.** `findLinks(text, logicalText?)` pairs the fragments with the URLs in that text by checking, not guessing:

- a fragment counts only when nothing but the row terminator follows it (`\n`, or herdr's own `\r\n`);
- it must be a prefix of exactly ONE URL there — duplicates of the same URL (a typed command and its output, which is how this was found) count as one;
- the URL is adopted only when its remainder is found character for character in the following rows, so a complete URL that merely prefixes a longer one elsewhere in the scrollback keeps its own href.

Adopted URLs give every fragment the same full href, so the whole URL is one tap target across however many rows it occupies. Fragments keep their own characters — this changes hrefs and adds ranges, never the text. Anything that fails a check keeps today's behaviour exactly, which is what the existing "stops at a newline" case still pins.

Two seams worth pointing at for review:

- `readLogicalText` is the first OPTIONAL method on `MuxAdapter`. Both decorators carry it the way `muxDataFields` carries `logo` (a `muxOptionalMethods` sibling, so an adapter that cannot unwrap never comes out of a decorator appearing to offer it), and `bridge/beacon/decorate.test.ts` names it in its contract fixture with cases for presence, absence and pass-through.
- herdr terminates a row with `\r` — that is why the fragment's href above is 103 rather than 120 — so both the gate and the repair treat `\r\n` as the row terminator.

## Coverage

- `web/src/lib/links.test.ts` — the repair rules: a fragment not at a line end, two different candidates, an unverified continuation, a three-row URL, CR-terminated rows, and the same URL appearing twice.
- `web/src/components/ansi-output.test.tsx` — the rendered mirror: every fragment of the wrapped URL carries the whole href, the rows are still exactly what the terminal printed, and the no-`logicalText` fallback is unchanged.
- `bridge/server.test.ts` — `hasSplitUrl` (styling, padding, CR, trailing prose, no URL), `paneReadResponse` with and without the field, and `readPane` issuing the extra read only when the grid shows a split.
- Checks: `bun run lint`, root and web `typecheck`, `bun test ./cli` (1405 pass, 0 fail), and the bridge files this touches — `bun test ./bridge/server.test.ts ./bridge/beacon/decorate.test.ts` (241 pass, 0 fail) — all pass.

Two things about running these locally, both pre-existing and neither related to the branch, in case CI sees them too: the web suite reports 150 failures here with an identical count before and after (localStorage under this Bun + jsdom combination), and a whole-directory `bun test ./bridge` never terminates on this machine because `bridge/stt/codex.test.ts` hangs — it hangs the same way at the release commit in a clean worktree, which is why the affected files were run directly.

## Changelog

Per CONTRIBUTING (from a fork: no version bumps), a line for `## [Unreleased]`:

> **A URL the pane wrapped is one whole link again.** The mirror renders the grid, so a URL longer than the pane was cut at the column edge — only its first fragment became a link, and that link opened a truncated URL. The bridge now fetches the same rows with soft wraps undone when the grid shows a split, and the mirror gives every fragment of that URL the href of the whole URL.

### DIFF
--- a/HERDR_API.md
+++ b/HERDR_API.md
@@ -55,6 +55,14 @@ the socket assumptions behind the design in [`ARCHITECTURE.md`](./ARCHITECTURE.m
     visible grid, and its renderer hard-wraps prose at the pane width — there are no soft-wrapped
     rows for the unwrap to merge. It only differs on scrollback-accumulating panes (shells: one
     probe measured 199 → 188 lines with logical lines up to 222 cols re-joined).
+  - **`recent_unwrapped` is how a wrapped URL gets repaired.** The mirror renders the grid, so a URL
+    longer than the pane arrives cut at the column edge and the client can only link the first
+    fragment — with a truncated href (`web/src/lib/links.ts`). The bridge asks for `recent_unwrapped`
+    when the grid it already has shows a URL touching a line end, strips the SGR and sends it as
+    `PaneReadResponse.logicalText`, which the client uses to give every fragment of that URL the href
+    of the whole URL. Same `lines` and same `ansi` format as the mirror read above, deliberately: the
+    observation that keeps this read from harvesting is the one the mirror read rests on too, and the
+    repair is only ever one extra read, on a pane that is showing the split.
   - **A `recent` text read can scroll the pane it reads.** Herdr's agent-automation docs state that
     for an idle, recognized agent at the bottom of its transcript, `recent` / `recent_unwrapped`
     reads "automatically use the agent's mouse-scroll interface" when `lines` asks for more than the
--- a/bridge/beacon/decorate.test.ts
+++ b/bridge/beacon/decorate.test.ts
@@ -5,7 +5,8 @@ import { withAgentHints } from "./hint.ts";
 import { fakeBeaconReader, FAKE_BEACON_NOW, type FakeBeacon } from "./fake.ts";
 import { BEACON_SCHEMA_VERSION, type BeaconMarker, type BeaconStatus } from "./types.ts";
 import { declareCapabilities, type MuxCapability } from "../mux/capabilities.ts";
-import {
+import {type MuxOutcome,
+  
   muxAck,
   muxGone,
   muxOk,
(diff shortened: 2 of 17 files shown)


AltanS:
Thanks @thelinuxlich, this is a sharp fix, and it lands. I keep your commit as it is, with your name on it, and one commit of mine sits on top. It also fixes a wrap that falls just after a `.` or `_`, which OAuth links hit all the time. It asks herdr for the unwrapped rows only when the next row can really continue the link. It also corrects the `readLogicalText` comment, which named `text` as the safe format when `ansi` is the one that does not scroll the pane. You get credit in the changelog, and GitHub closes the PR when main goes out.

PR #221, opened 2026-09-15
fix(mirror): muse panes render natively in light instead of inverting to ~2:1

### LINKED ISSUE #220: Light-mode mirror unreadable for Muse panes (light-adapted palette + inversion)
## Summary

Muse panes are unreadable in Collie's light mode: body text renders as light grey on white (~1.6:1 contrast) while saturated accents (links, statusline) look fine. Root cause is a three-layer interaction — Muse adapts its palette to the terminal background it probes via OSC 11, and Collie's light-mode inversion assumes dark-theme agent output (ADR 0002). Dark mode is not an available workaround for this reporter (visual difficulty), and with these bytes dark mode would show dark-on-dark anyway.

## Setup

- Collie bridge 1.5.3 (PWA served from it), Android phone
- Herdr 0.9.0 with `solarized-light` theme
- Muse Code 1.2.1 in the pane

Repro: run `muse` in a Herdr pane under a light Herdr theme, open the pane in Collie with light appearance.

## Root cause (verified with captures)

1. **Muse probes the terminal background.** At startup it emits `OSC 10;?`, `OSC 11;?`, and `OSC 4;0..15;?` (captured from a live TUI under a pty).
2. **Herdr answers truthfully.** The embedded Ghostty core answers color queries from its configured colors (`stream_terminal.zig` → `writeXtermColorReport`), so a light Herdr theme reports a light background.
3. **Muse emits a light-adapted palette.** Same prompt, different OSC 11 answers, captured bytes differ:
   - Light terminal: body `38;2;56;58;66`, secondary `38;2;121;122;128`, hints `38;2;175;176;180`, accents `38;2;163;45;161` / `38;2;193;132;1`, all with transparent bg (`49`).
   - Dark terminal: body `38;2;117;120;129`, secondary `38;2;100;103;110`, hints `38;2;82;84;90`, accents brighter.
4. **Herdr serializes faithfully.** `pane.read format:"ansi"` (Ghostty VT formatter) passes explicit truecolor through as SGR and emits default-styled cells bare — no theme-dependent remapping.
5. **Collie inverts.** The light-mode mirror applies `invert(1) hue-rotate(180deg)` over a dark ground. `(56,58,66)` becomes `~(199,197,189)` on near-white: about **1.6:1**. Saturated accents survive via hue-rotate, which is exactly the observed symptom (pink/orange fine, body washed out).

I also checked Collie's SGR parser (`web/src/lib/ansi.ts` at 1.5.3): it handles these sequences correctly, and the harness registry has no `muse` adapter (exact-match only, so Muse gets the raw mirror — correct behavior, just unvalided palette). Nothing is mis-parsed anywhere; ADR 0002's "agents emit dark-theme colors" assumption is what breaks, because Muse's emission follows the host theme.

Note this also means a Herdr **dark** theme would only soften the bug (body `117,120,129` inverts to ~2.8:1 — still poor), not fix it.

## Suggested fixes

1. **Muse-specific (small, precedented):** add a `muse` harness adapter with a display transform — the fg analogue of the existing `mobileTransparentBg` hint used for codex/omp fills. Options within it: skip inversion for Muse panes whose palette reads light, or lift low-luminance fg spans pre-inversion.
2. **General (helps any dim/adaptive palette):** make light-mode inversion contrast-safe, e.g. remap sub-threshold-luminance foregrounds before inverting, so future adaptive agents don't each need an adapter to be readable.

Happy to test builds against the live Muse pane, provide full byte captures of either palette, or attach a screenshot of the current rendering.

jpcarranza94:
Muse's mid-tone palette reads raw on either ground, while the light-theme inversion drops its body text to ~2:1 light-grey-on-white. Muse panes now skip the inversion on the page ground, with bright foregrounds resolved dark through a light-gated custom property. Dark rendering is pixel-identical; other agents are untouched.

Fixes #220. The full evidence chain (live palette captures across background answers, contrast figures, rejected alternatives) is in `.adr/0046-muse-panes-render-natively.md`, included here.

**Base branch** — see [CONTRIBUTING.md](../CONTRIBUTING.md#base-branch):

- [x] Opened against `main` — there is one line of development, and `main` is it

**Checks**

- [x] `bun run typecheck` (root **and** `web/`), `bun run lint`, `bun test ./bridge ./cli ./scripts`, `cd web && bun run test` all pass
- [x] CHANGELOG entry added under `## [Unreleased]` … (fork PR: exempt — suggested entry below, maintainer picks the version)

Note on the bridge suite: it carries 1 failure on clean `main` (`resolveBlobPath > resolves candidate contained in sibling blobs directory`), reproduced before any change; this PR touches `web/` + `.adr/` only.

Suggested CHANGELOG entry, under `### Fixed`:

- **Muse panes render natively in light mode.** The mirror no longer inverts Muse's mid-tone palette into ~2:1 grey-on-white; the pane sits on the page ground with only bright foregrounds resolved dark, and dark rendering is unchanged. (#220)


jpcarranza94:
Verification notes for the reviewer, since the interesting half isn't in the diff:

- **Live bytes, not just captures.** I pulled the real pane (`herdr pane read wM:p1 --format ansi`) and confirmed the agent string is exactly `muse`, the palette is the light row (body `56,58,66`), and there are zero bg fills and zero indexed foregrounds in 9KB of transcript. The 0.6 threshold splits the live corpus with margin (brightest non-white observed: `216,216,218` rules at L≈0.69 — muted, so they resolve through the chrome var, not the mark).
- **Headless render.** Live bytes through the real AgentChat in Chromium, light + dark + 390px viewport: light is desktop-terminal-identical (body 10.4:1), dark is pixel-identical to the old path, reflow stable. Screenshots verified by eye.
- **Heads-up, out of scope:** HERDR_API.md's 'Herdr answers neither OSC 10/11' probe (2026-07-29) looks stale at 0.9.0 — `write_pty` is wired in `src/pane/terminal.rs` now, and the live palette proves queries are answered from the theme. ADR 0046 notes this; the doc itself may want a re-probe.

PR #222, opened 2026-09-15
fix(triage): show unread idle completions from Herdr 0.9

magoz:
## Summary

Restore **Ready · unseen** for completed agents reported as `idle` by Herdr 0.9. The section was not removed: its classifier required literal `done`, while Herdr's terminal client derives `done` from its own acknowledgements. Collie's API snapshot can therefore put the same completion in Recent.

- Recognise settled `idle` and `done` agents using Collie's existing `lastActiveAt > lastSeenAt` read receipts; exclude shells.
- Preserve first-sighting equality, older-bridge fallback, and shared clearing when a pane is opened in Collie.
- Reset activity on an observed agent exit before reseeding a surviving shell, so a new idle agent cannot inherit unread work.
- Add classifier, dashboard, folded-switcher and real state-engine/activity-ledger regression coverage; document the Herdr 0.9 distinction.

No sorting changes. The separate priority-order PR #223 can land independently.

## Visual proof

Actual built app with synthetic data derived from the repository's API fixtures, Chromium 390×844. **Finished while away** has API status `idle` and now appears above the working agent, under Ready · unseen.

Unread idle completion above working agents

The image uses its existing hosted URL; no screenshot files or screenshot commits are included in this PR’s branch.

Capture source: base `96e451a4c2dbb77978a69bfc036dc5fd84cfb535` + frontend patch digest `1efd8c6f92c0c6ce931a8df3fc5368587f3115e5`. The final patch `252b2105975b9e7b8ee058b47b8b0d357ad810af` adds only backend lifecycle handling/tests and its ADR note; the captured frontend is unchanged. Tested implementation: `2be89b59d3529faf59e596113ae243e56f35730b`. No private conversation screenshots or live-pane actions are included.

## Validation

- New frontend regression cases failed before the classifier fix; lifecycle regression failed without the removal reset.
- Frozen root/web installs, both typechecks, full-tree `bun run lint`, and `bun run build` — passed.
- `cd web && bun run test` — 224 files, 6,051 passed, 30 existing TODOs.
- `bun test ./bridge ./cli ./scripts` — 4,829 passed against the final implementation.
- `git diff --check` — passed.
- CI `typecheck + tests` and full Playwright Chromium/WebKit matrix — passed on the current head. Live multi-device completion/clearing was not run; production remains untouched.

## Review

[...]

PR #223, opened 2026-09-15
fix(triage): restore latest-state-change ordering

magoz:
## Summary

Restore latest-state-change ordering inside the triage sections after `6e8eeafc` switched them to workspace/tab order. The switcher's Recent section currently shows older activity above newer activity.

- Keep Needs you → Ready · unseen → Working → Recent.
- Sort each section by descending `lastActiveAt`, not `lastSeenAt`; opening a pane does not promote it.
- Reverse only Recent's time comparator when requested, preserving ties and missing-timestamp fallback.
- Leave the bridge's positional order and workspace/tab navigation unchanged.

This is independent of the unread-idle completion fix #222; it does not change bucket classification.

## Visual proof

Actual built app with synthetic data derived from the repository's API fixtures, Chromium 820×1180. Recent shows **Newer completion** above **Older completion**, although the fixture's workspace order is the opposite.

Switcher ordered by latest state change

The image uses its existing hosted URL; no screenshot files or screenshot commits are included in this PR’s branch.

Capture source: base `96e451a4c2dbb77978a69bfc036dc5fd84cfb535` + patch digest `113935464136c62a955cde65f719952fa6490f33`, identical to the implementation committed as `ce1419d4be6fe32322eaed344a0faea289fb3dc7`. No live panes or production service involved.

## Validation

- Frozen root/web installs, both typechecks, full-tree `bun run lint`, and `bun run build` — passed.
- `cd web && bun run test` — 224 files, 6,049 passed, 30 existing TODOs.
- `bun test ./bridge ./cli ./scripts` — 4,828 passed on the identical backend/CLI/scripts tree and dependencies in the companion checkout; reused unchanged evidence.
- `git diff --check` — passed.
- CI `typecheck + tests` and full Playwright Chromium/WebKit matrix — passed on the current head. Browser matrix was not rerun locally.

## Review

Ready gate completed at `ce1419d4be6fe32322eaed344a0faea289fb3dc7` against `main@96e451a4c2dbb77978a69bfc036dc5fd84cfb535` (patch digest `113935464136c62a955cde65f719952fa6490f33`). Existing local tests, builds, Spec review and inspected visual evidence were reused because the implementation is unchanged. Fresh independent Standards and Knowledge reviews completed; full CI is green at this exact head.

Standards, Spec and Knowledge reviews passed without findings. No unresolved review findings remain.

## Remaining risks

This restores Herdr's **priority then latest state change** approach, not exact cross-client parity. Collie observes `lastActiveAt` via polling; Herdr's terminal client uses its internal `state_change_seq`. Unobserved/pre-start transitions, same-millisecond timestamps, clock adjustments, and independent read receipts can produce different orders. Incoming order breaks ties deterministically.


AltanS:
Thanks @magoz, this is careful work, and I am not taking it. 1.9.0 moved every section to the multiplexer's own order on purpose: a row only moves when it changes section, so the pane you reach for is still where it was a minute ago. Ordering by latest state change is what Herdr's list does, and it is exactly the movement under the thumb that decision removed. Your diagnosis of the gap is right, and the part that matters, a finished agent stuck in Recent, landed this morning with #222, with your authorship kept and credit in the changelog. I am closing this one, and the ordering stays a settled decision rather than an open one.

PR #229, opened 2026-09-15
fix(mirror): ground native light mirror on probed reference bg

jpcarranza94:
Follow-up to #221 (shipped in 1.9.1): the native light ground it chose (`#f5f5f5`, the page colour) sits 10 steps below the ground Muse's light palette was authored against.

I probed Herdr 0.9.0's light background live via OSC 11 on a scratch pane: `rgb:ffff/fbfb/f8f8` (`#fffbf8`). On page ground, Muse's authored fills compress past visibility — the 236 prompt fill renders at 1.08:1 and the prompt block melts into the page (phone photo on #220's follow-up thread). Matching the probed reference restores the authored 1.15 edge and lifts every tone with it:

| span | on `#f5f5f5` | on `#fffbf8` |
| --- | --- | --- |
| body 56,58,66 | 10.4 | 11.0 |
| secondary 121,122,128 | 3.9 | 4.2 |
| hints 175,176,180 | 2.0 | 2.1 |
| near-white mark | 18.2 | 19.2 |
| muted rule | 6.0 | 6.4 |
| prompt-fill edge | 1.08 | 1.15 |

The whisper seam against the page is deliberate — fidelity to the agent's own screen beats seamlessness. Dark untouched. Verified in a headless render (computed ground `rgb(255,251,248)`, prompt block reads as a distinct bar) and running live on my own install.

- ADR 0047 gains an amendment section with the probe and revised figures
- Ground literal pinned by a new `MUSE_MIRROR` test; native-mirror className pin updated


AltanS:
Thanks @jpcarranza94, the probed ground is right and it shipped in 1.10.0. Your commit is on main with your authorship, and I added one fold on top that leaves the ADR's Trail line as written and scopes "pixel-identical" to Herdr 0.9.0's default light theme. You are credited in the changelog, so I am closing this.

PR #230, opened 2026-09-15
fix(mirror): trim Muse row chrome so hard breaks read as soft wraps

jpcarranza94:
Follow-up to #221, from a phone photo: Muse hard-wraps prose at the PTY width (116 cols here) and opens every row with a 2-column gutter, padding each row to full width. On a desktop terminal both are invisible structure; on a ~55-column phone each hard row soft-wraps and the gutter lands mid-paragraph as a stray indent while the padding becomes blank stub lines — the "terminal didn't resize" look.

The Muse display pass now trims both per row (`trimMuseRowChrome`, ahead of the bright-foreground marks), so hard breaks read exactly like soft wraps. Verified against the photo's exact scrollback bytes in a headless render (gutter indents gone, paragraph reads as clean soft-wrapped text).

Signatures are narrow on purpose, all grounded in a 1000-row live capture:

- gutter: exactly two spaces, must carry a foreground (grey prose gutter; bold-body emphasized continuations), no background, no underline/strike
- padding: trailing whitespace-only spans without background or lines — fills (prompt, diff) keep theirs and stay full-width
- wider grey leads (a 7-wide table field was observed) and bare leads stay byte-faithful
- blank rows collapse to zero segments (same blank render, ~116 fewer space chars in the DOM each)

Gated to Muse panes like the marks: other agents' grammars may depend on exact row text (a Codex wrapped card is recognised BY its wrap). Autolink repair is unaffected (prefix matching on URL text), and `table-run` already trims row ends before detection.

- 12 new `trimMuseRowChrome` tests + 2 `buildBlocks` wiring tests (muse trims, codex identical)
- ADR 0047 gains a follow-up section


AltanS:
Thanks @jpcarranza94, the row trim lands exactly as scoped and shipped in 1.10.0. Your commit is on main with your authorship, plus one fold that moves the new helpers above `decorateMuseDisplay` so its doc comment stays attached. You are credited in the changelog, so I am closing this.

PR #231, opened 2026-09-15
fix(update): an update that finds its target already staged closes the record it opened

foreverrrree:
## Summary

`collie update` on a staged checkout opens the run record (`update.json` → `staging`) at `beginStaging`, before the fetch. `withStagingRecord` closes it on every **failing** exit. One exit inside that window is not a failure: the target is already what `current` points at, complete and live, so the run prints `already current — vX is staged and current points at it.` and returns `EXIT.OK` — and the record is left at `staging` under a pid that has exited.

Ten minutes later the staleness rule turns that into `interrupted`, and the phone's Updates page reports the updater as gone mid-flight, about an install that is exactly what it should be. `collie update --status` says the same:

```
interrupted — the updater is gone and 1.9.1 was mid-flight
  reason: the updater (pid 714911) is gone and the run has not moved since …
```

## How it is reached

The branch is decided from `current`, not from the manifest of the version we are running — its own comment names the case: an operator running the old binary by hand right after `current` flipped. That is what happened here on 2026-09-16: the phone's update to 1.9.1 completed (`✓ updated to 1.9.1` in the journal), and a `bin/collie update` from the pre-staging 1.9.0 checkout ran a few minutes later, found v1.9.1 staged and live, said "already current", and left the record open. The phone then showed the interrupted state until the file was fixed by hand.

## The change

- `abandonStaging` takes a `reason` (default unchanged: `STAGING_GAVE_UP`).
- The already-staged exit calls it with `ALREADY_STAGED` before returning `EXIT.OK`, so the record goes back to `idle` carrying a reason that says what happened — `abort` is the state machine's own word for a staging that moved nothing.

Two lines of behaviour, one constant, one test. The test reproduces the branch by seeding a stale root manifest (`0.32.0`) under a `current → v1.0.0` that is complete, runs `update --major`, and pins the record at `idle` with no `worktree add`. It fails on `main` with `Received: "staging"`.

## Validation

- `bun test ./cli` — 1460 tests, green (307 in `update.test.ts`)
- `bun run typecheck` (root)
- `oxlint` over the two files, clean. (The full-tree `bun run lint` cannot run on this host because the staged `versions/` worktree sits inside the checkout and oxlint registers the vendored plugin twice — that is my layout, not this change.)

Fork PR: no version bump, no CHANGELOG line. If you'd like one in my words:

> **A second `collie update` beside a live one no longer reads as interrupted.** An update that finds its target already staged and live returns the run record to idle instead of leaving it at `staging` under an exited pid, so the phone stops reporting a mid-flight updater about an install that is fine. Thanks @foreverrrree.


AltanS:
Thanks @foreverrrree, your diagnosis was right and the fix shipped in 1.10.0. Your commit is on main with your authorship; on top, the already-live check now runs before the staging record opens and before the Bun probe from #232, so such a run writes no record at all and a waiting phone never reads it as failed. You are credited in the changelog, so I am closing this.

PR #232, opened 2026-09-16
Fix managed Bun update and build isolation

en-ver:
## Summary

- Probe the resolved Bun executable before managed or staged updates mutate a checkout, and hand the exact verified path to the update child.
- Isolate compiler scratch in owned canonical `bin` directories, preserve legacy generated sidecars while detecting new sidecars, and publish default `build:cli` output only after verification and cleanup.
- Route release and AUR compile paths through the explicit build output contract.
- Add hermetic direct, fork-guard, detached-shell updater, build-publication, cleanup, and hostile-path regression coverage.

## Validation

- `bun test ./cli/build.test.ts ./cli/sys.test.ts ./scripts/build-cli.test.ts` (68 passed)
- `bun test ./cli/update-check.test.ts ./cli/update.test.ts` (226 passed)
- `bun test ./scripts/check-flake-bun.test.ts` (5 passed)
- Root and web typechecks; `bun run lint`; affected shell syntax checks; `scripts/check-version.sh`
- AUR PKGBUILD static suite, release-workflow YAML parse, and `git diff --check`

## Environment limits

The full `scripts/collie-cli.test.sh` was not rerun because it rebuilds `bin/collie`, and this macOS host cannot meet its Linux `systemd-run` expectation. Linux service-manager, AUR VM/container, and release-workflow execution remain unrun; these environment limits (including known macOS baseline limitations) are not presented as patch failures.


AltanS:
Another great contribution, thanks @en-ver. Both halves are sound and shipped in 1.10.0, and the macOS payload built cleanly in your compile sandbox. Your commit is on main with your authorship, plus one fold that restores the shell-suite guard that the unit names no interpreter and drops the backticks from the fake `systemd-run` heredoc comment. The changelog credits you and names the new red state for an unrunnable Bun and the real-directory rule for `bin/`, so I am closing this.

PR #235, opened 2026-09-17
fix(lifecycle): fall back to `launchctl load -w` when bootstrap persi…

cmal:
## Problem
On some macOS installs, `launchctl bootstrap` fails persistently with `Bootstrap failed: 5: Input/output error` even when the plist is valid, the binary exists and is executable, and all required directories are present. The legacy `launchctl load -w` command succeeds on the same machine.

Because `cmdRestart` (called after every self-update) runs `cmdStop` (`launchctl bootout`) followed by `cmdStart` (`launchctl bootstrap` with 3 retries), an update on such a machine always leaves the bridge down.

## Fix
After the three `bootstrap` retries fail, try `launchctl load -w <plist>` once. If it succeeds, the service stays under launchd (KeepAlive, RunAtLoad). If it also fails, the existing unsupervised fallback is unchanged.

## Tests
- Updated the existing "degrades to unsupervised" test to also mock `launchctl load` as failing.
- Added a new test verifying that when `bootstrap` fails 3× but `load -w` succeeds, the bridge reports `bridge started (launchd/load: ...)`.
- All 60 lifecycle tests pass.


AltanS:
Thanks @cmal, the bug is real and your reproduction is clear, but I cannot take this shape. `launchctl load` takes no domain, so it loads into the caller's session domain: `gui/<uid>` from a console shell, `user/<uid>` over SSH. Stop and status address `gui/<uid>/<label>` only. So on the machine this is written for, a load-started agent survives `collie stop`, status reports it as not loaded, and the next start puts a second bridge on the same port. Every launchd spelling in start needs a matching one in stop and in status, which is a larger change than this. I also want the cause pinned first, because a persistent `Bootstrap failed: 5` with a valid plist is what launchd returns when the label is already registered in another domain. Could you run `launchctl print gui/$(id -u)/herdr.collie`, `launchctl print user/$(id -u)/herdr.collie` and `launchctl print-disabled user/$(id -u)` while bootstrap is failing, and paste all three? If that shows a stale registration, the fix belongs in stop, and your authorship stays on whatever lands.

PR #239, opened 2026-09-17
feat(release): add Intel Mac support with a macos-x64 payload

ryands17:
Releases ship three payloads today: linux-x64, linux-arm64 and macos-arm64. An Intel Mac gets no
binary, and `collie update` points it at a source build. This PR adds the fourth payload.

## What changes

- `release.yml`: a `macos-x64` matrix row on `macos-26-intel`, target `bun-darwin-x64`. It is a
  standard runner, so free on a public repository, and the last Intel image GitHub will offer,
  since macOS 26 is Apple's last Intel release. The comment says to drop the row when it goes.
- `flake.nix`: `x86_64-darwin` joins the systems, with the pinned Bun 1.4.1 `bun-darwin-x64.zip`
  source that `scripts/upstream-bun.sh` realises on that runner. `flake.lock` does not move.
- `scripts/refresh-packages.ts`: a `darwin-x64` row in the Nix table. Without it, the packaging
  job after the release refuses the manifest as carrying an unmapped platform.
- `packaging/nix/collie.nix`: maps `x86_64-darwin` to that key, and refuses by name while
  `sources.json` still records a release without the payload.
- Tests and docs: the packaging and flake-bun tests carry four platforms; `docs/install.md`
  names four flake systems.

`scripts/install.sh` and `platformId` in `cli/update.ts` already answer `macos-x64`, so an Intel
Mac can use the installer and `collie update` as soon as the artifact exists.

## Verified

- Root typecheck and oxlint pass.
- `scripts/check-flake-bun.test.ts` passes.
- `packaging/refresh.test.sh` shows the same three failures on `main` on my Mac, so those are
  pre-existing and unrelated.
- Not evaluated: the flake on `x86_64-darwin`. Nix is not installed here; the release run is the
  first proof.

## Not in this PR

No CHANGELOG line and no version bump, per the fork rule. Suggested Unreleased line, under
`### Added`: **Releases ship an Intel Mac binary.** A `macos-x64` payload joins the release,
built on GitHub's Intel runner.


AltanS:
Thanks @ryands17, this is the right shape and it lands. Your commit is kept as you wrote it, with your authorship, and I checked the pinned `bun-darwin-x64.zip` hash byte for byte against the published Bun 1.4.1 archive before I merged it. On top I added the changelog line that credits you, and a source for the one claim the comment made without one, that every Intel Mac on macOS 26 has AVX2, which Apple's own compatibility list backs because it lists no Intel model older than 2019. The piece I cannot supply from here is proof on real hardware, since nobody on this side owns an Intel Mac, so I am cutting a throwaway prerelease tag that builds the payload and nothing else. If you are willing to download that tarball and run `collie doctor` on your Mac, I will post the link here, and that is the last evidence I want before the release.

AltanS:
@ryands17, I have to take this back out, and it is not your work that is wrong. The rehearsal tag built the three existing payloads green and could not build the fourth, for two reasons that both sit upstream of this repository. `nix develop` on `macos-26-intel` answered "Nixpkgs 26.11 has dropped support for x86_64-darwin", and at the revision the flake pins every package throws, not only the slow ones, so `scripts/upstream-bun.sh` dies the same way because it realises `#bun.src` from that same flake. In the same run the installer refused the machine outright, with "Determinate Nix Installer no longer supports macOS on Intel". Only the nixpkgs 26.05 branch still carries that platform, and pinning a second nixpkgs for one matrix row would mean one release built with two toolchains, which is the property the flake exists to hold. So I reverted the merge and dropped its changelog line, since it never shipped, and the fourth matrix row is commented out again with the reason written beside it. Your commit and your authorship stay in the history, and if either blocker moves, this comes back as your commit.

PR #241, opened 2026-09-17
fix(web): a per-pane don't-invert override for light-themed agent panes

### LINKED ISSUE #240: opencode panes need native mirror bit: light theme inverts to black background
What happened, and what did you expect?

opencode panes show a black background in Collie light theme when opencode uses a light-authored theme. Expected readable light ground like Herdr TUI shows.

Setup:
- Collie 1.10.1+166cabd, Herdr 0.9.0, opencode 1.18.31, macOS arm64
- Herdr theme `one-light`, `auto_switch = false`
- opencode `tui.json` theme `github-light`, custom theme with `background #ffffff` for both dark and light variants, `backgroundPanel #f6f8fa`, `backgroundElement #eaeef2`
- Collie app theme light

Repro:
1. Run opencode in a Herdr pane with the light theme above. Herdr TUI renders white ground correctly.
2. Open the same pane in Collie with light theme selected. Pane ground renders black.

Root cause hypothesis, from your own docs:
- ADR 0002 measures opencode at 100 percent truecolor, 0 percent themeable, so the light mirror invert (`invert(1) hue-rotate(180deg)`) is the only thing making dark-authored output readable.
- ADR 0002 Known limitation already calls out agent-on-light as unreadable in both Collie themes, with fix noted as per-pane do-not-invert.
- ADR 0047 then implements exactly that for Muse via `rendersNativeMirror` / `NATIVE_MIRROR_AGENTS` in `web/src/lib/harness/muse/display.ts`, because Muse probes OSC 10/11, adapts palette, and uses transparent backgrounds.
- opencode is the same class now: Herdr 0.9.0 answers OSC 10/11 from its theme (ADR 0047 Context confirms the old no-answer probe is stale), so under Herdr `one-light` opencode probes light and emits its light variant (white ground). Collie then inverts white to black.
- `NATIVE_MIRROR_AGENTS` currently contains only `muse`. Request: evaluate opencode for the same bit, with opencode live PTY measurements per ADR 0047 follow-up rule (a second agent wanting the bit adds one exact string with its own measurements).

Workaround on my side is to run opencode dark-authored or `system` under Herdr, but that sacrifices the light Herdr TUI appearance. Happy to provide a live `herdr pane read` capture or test a preview build.

Related prior art in repo: issue #220 (Muse), #144 / #224 (Codex black bars), ADR 0002, ADR 0047.

waynehoover:
Opencode panes show a black background under Collie's light theme when opencode runs a light-authored theme. Fixes #240.

Root cause: opencode emits effectively 100% truecolor (the ADR 0002 measurement, re-confirmed below), so the light mirror's wholesale inversion turns its white ground black. This is the same class ADR 0047 fixed for Muse, and the second-agent case the native-mirror bit was reserved for.

Measurements (live `herdr pane read format:ansi`, opencode 1.18.31 with a light theme under Herdr one-light, 60 rows, 706 SGR sequences, zero palette-indexed):
- grounds `rgb(255,255,255)`, `rgb(246,248,250)`, `rgb(234,238,242)`, all within 1.13:1 of the native `#fffbf8` ground
- body `rgb(31,35,40)` at 15.35:1, muted at 5.94:1, link at 5.05:1, all readable raw
- the only bright tone is white foreground, and every non-blank span carrying it also carries an explicit background (badges and chips), so the fg+bg exemption keeps those pairs as authored
- no styled 2-space lead observed, but absence in 60 rows proves nothing (see below)

What changed:
- `NATIVE_MIRROR_AGENTS` gains `opencode`: native ground, no inversion, bright-foreground marks, raw-terminal keeps native rendering.
- `trimMuseRowChrome` stays Muse-only behind a new `trimsRowChrome` predicate. A probe through the real parser showed a styled 2-space lead (indentation) would otherwise be stripped on opencode rows, removing visible text in either theme. This was caught in review and is pinned by a test.
- Tests: membership predicate for both agents plus near-miss exactness, native ground and observed corpus shapes for opencode, find-match without the cancelling filter for both, trim preservation on opencode rows, conservative-gating text equality in `blocks.test.ts`.

Deliberately not included:
- No new adapter, so the adapter-conformance and live-keystroke verification bars do not apply. No fixture file: the agy fail-closed contract covers every `.txt` in the fixtures directory, and the synthetic tests already run the real `parseAnsi` to `splitLines` to `buildBlocks` to render pipeline on the observed shapes, the same precedent as the Muse change.
- Versions and CHANGELOG untouched per the fork guidance. Suggested line for `## [Unreleased]` under `### Fixed`: **Opencode panes render natively in the light theme.**

Checks: root typecheck, web typecheck, oxlint, and the full web vitest suite (231 files, 6683 passed) are green. The backend suite carries one pre-existing failure (`bridge/journal/pi.test.ts` blob-path assertion, fails on clean main too, macOS TMPDIR symlink artifact of the sandbox). Browser e2e was not run locally.


AltanS:
Thanks @waynehoover, this is careful work with the right method, and I need one more measurement. ADR 0047 granted the native-mirror bit to Muse after measuring all three background answers, light, dark and none. This measures one: a light-authored opencode theme under a light Herdr theme. Membership in `NATIVE_MIRROR_AGENTS` is a fact about the agent, not the theme, so the same bit would render a dark-authored opencode pane raw inside a light Collie, and opencode's default theme is dark. Today that pane inverts and reads. Could you take the same capture against the default dark theme and post the contrast figures? If it reads raw, this lands as it stands. If not, the right seam is the per-pane do-not-invert override ADR 0002 named, not the agent set. The `trimsRowChrome` split is correct either way and I will keep it.

waynehoover:
Measured it, and it does not read raw — so this lands as the second of your two branches, not the first. Thanks for holding the gate; the bit would have been wrong.

## The capture

opencode 1.18.31 under a pty answering OSC 10/11, **isolated XDG dirs so no persisted theme applies**, 100 SGR truecolor sequences per answer. WCAG against the native `#fffbf8` ground, raw versus what inversion renders today:

| default theme, **dark** answer | raw | inverted (today) |
| --- | --- | --- |
| body `rgb(238,238,238)` | **1.13:1** | 17.32:1 |
| white `rgb(255,255,255)` | **1.03:1** | 19.26:1 |
| link `rgb(92,156,245)` | 2.71:1 | 4.43:1 |
| muted `rgb(128,128,128)` | 3.84:1 | 3.67:1 |

| default theme, **light** answer | raw | inverted |
| --- | --- | --- |
| body `rgb(26,26,26)` | 16.91:1 | 1.16:1 |

opencode is Muse's inverse. 0047 could grant Muse the bit because every Muse answer is mid-tone and reads raw in all three. opencode's dark answer is a conventional near-white-on-near-black theme: raw on a light ground the body vanishes at 1.13:1, while today's inversion carries it at 17.32:1. Here the inversion is preserving the agent's contrast, not spending it.

Worth noting the default theme *does* answer the background query — under a light answer it emits `rgb(26,26,26)` and would read raw. But membership is a fact about the agent, so it cannot say "only when this pane happens to be light", and Herdr's theme is independent of Collie's: a dark Herdr pane in a light Collie is the case that breaks. That is your per-pane override, exactly as 0002 reserves it.

## Why my first measurement missed this

It was not a light-authored *theme* I happened to pick — opencode persists the selected theme in its own state, and this machine had `"theme": "github-light"` set there. So the capture reflected my own setup rather than what a new user sees, and both background answers returned the same light palette, which is why the dark case never showed itself. That is a methodology failure on my part, and the isolated-XDG run is what exposed it.

## What this PR is now

Only the `trimsRowChrome` split, which you said you would keep and which is right regardless: the row-chrome trim is Muse's grammar, not the mirror's, so it is gated separately rather than riding on the native-mirror predicate. opencode is out of `NATIVE_MIRROR_AGENTS`, and the measurement is recorded in `display.ts` so the next person does not re-litigate it.

One thing I want to flag rather than hide: with Muse the only native agent, `trimsRowChrome`'s false arm is now **unreachable**. I kept it because it is the seam the next native agent needs and you asked for it, but it is dead code today and I would rather you know that than discover it. Retitled the PR to match what it actually does.

## For the follow-up

@waynehoover reports the same unreadable-pane symptom on **codex** panes sometimes, which is a second data point for the per-pane override being the right seam rather than a growing agent set. Happy to take that as a separate issue and measure codex the same way if useful.

Tests, typecheck and `oxlint --max-warnings 0` are clean (231 files, 6679 tests).

waynehoover:
Built the override rather than leaving it as a follow-up, so this PR now fixes the reported symptom instead of only explaining it. Two commits: the measurement and the `trimsRowChrome` split, then the seam.

**The override** is per-pane, tri-state, and wins in both directions. `undefined` means "let the agent bit decide" and is deliberately not the same as `false` — without that distinction the toggle would have to lie about a Muse pane, showing off while the pane rendered natively anyway. Choosing the agent's own answer *clears* the entry rather than pinning it, so a pane does not freeze on today's answer if 0047's set changes under it later.

**Storage** is one localStorage entry per pane keyed by scope+pane, the shape `lib/last-seen.ts` uses. Scope-keyed because two machines in a pack can both have a `w1:p1`; bounded at 32 because Herdr reuses pane ids as workspaces come and go, so an unbounded map would eventually apply a stale decision to an unrelated pane.

**`AnsiOutput` takes the resolved boolean, not a paneId.** It stays presentational and never reads storage; `AgentChat` owns the lookup, since it is already keyed by paneId and holds the scope.

One consequence worth calling out: this makes the `trimsRowChrome` split **load-bearing** rather than the dead-but-kept seam I flagged last time. An opted-in opencode pane is now genuinely native, and must be decorated *without* Muse's row-chrome trim. The false arm is reachable the moment anyone uses the toggle.

Also: codex gets the fix for free. @waynehoover reports the same symptom there, and nobody has to measure codex's palette or add it to a set — the operator can see which way their own pane points, and this predicate cannot.

`NATIVE_MIRROR_AGENTS` is still exactly `["muse"]`, unchanged from main.

### Two things I want your eyes on

1. **The i18n strings.** `settings.display.noInvert.{label,hint}` had to land in all 7 locales because they are typed from `en`. The de/es/ja/ko/zh/zh-TW versions are mine, not a native speaker's — please treat them as placeholders if you have better.

2. **Toggle placement.** I put it in the existing display-prefs sheet next to Raw terminal, which is the closest neighbour conceptually. But every other row there is a *global* pref and this one is per-pane, so the sheet now mixes scopes. If that bothers you I am happy to move it to a pane-scoped surface instead.

Gate is green: 232 files, 6690 tests, `tsc --noEmit`, and `oxlint --max-warnings 0` (the anti-slop rules pushed the storage module onto the per-key `last-seen` idiom, which is better than what I first wrote).

AltanS:
Thanks @waynehoover, the measurement settles it and the second branch is the one to build, which is what you built. Recording the capture in `display.ts` is where that belongs, so nobody re-opens this from a screenshot. On your two questions: leave the toggle where it is, because `DisplayPrefsContent` has exactly one caller, the composer dock, which is already opened from inside a single pane, and your label and hint name the pane themselves. The translations need one fix and not a rewrite, since zh and zh-TW call a pane 窗格 everywhere else in the catalog and ko calls it 창, so 面板 and 패널 are the wrong noun in those three, and de, es and ja are right as they stand. One thing worth knowing before you rebase, main has since grown a `grammars` flag on `buildBlocks` and `AnsiOutput` that turns the adapter off for raw terminal while a native agent keeps its display passes, which is what your last commit does by hand, so three files conflict and all three resolve by keeping `grammars` and adding `nativeMirror` beside it. Your authorship stays, the changelog line credits you, and the PR closes when the branch merges.

waynehoover:
Rebased onto main and applied the translation fix. Force-pushed; the three conflicts are resolved and the branch is clean against `7dfee360`.

## The rebase

Your prediction was exact: `agent-chat.tsx`, `ansi-output.tsx` and `lib/harness/index.ts`, and all three resolve by keeping `grammars` and adding `nativeMirror` beside it.

- `buildBlocks` now takes `ctx?: { agent?: string; grammars?: boolean; nativeMirror?: boolean }`. Your `grammars === false` adapter gate is untouched and still the thing that turns the adapter off for raw terminal; `nativeMirror` only reaches the native-mirror predicate below it, so the display passes stay on the other axis.
- `AnsiOutput` takes both props and threads both into `buildBlocks`, with both in the memo deps.
- The call site passes `grammars={grammarsOn}` and `nativeMirror={mirrorOverride}`.

One thing I did not resolve mechanically, because taking my side verbatim would have left a comment that is now false. My last commit gated the `agent` prop on the agent bit alone, and the comment explaining it said that folding `mirrorOverride` in would hand an opted-in codex pane its grammars back under raw terminal. With `grammars` on main that is no longer true: `grammars={grammarsOn}` is the authoritative adapter gate, so the adapter would not run whatever `agent` says. The gate is still right, so the code is unchanged, but I rewrote the comment to say why it is right *now* (the override does not need to travel that way, since `nativeMirror` carries it to the mirror directly and identity should keep tracking the agent rather than a display choice) and kept one sentence of the old reason as history, marked as history. Worth a look, since it is the one place where the merge was a judgement call rather than a union.

## The translations

Fixed the noun in the three you named, and only those. zh and zh-TW use 窗格 for a pane in 50 and 51 other strings against 4 and 3 uses of 面板; ko uses 창 in 58 against 3 of 패널. So the new label and hint were naming a different piece of UI than the one they act on. de, es and ja are untouched.

Two outliers are left alone because they predate this branch and are not mine to change, but you may want them at some point: `settings.zen.description` in zh says 面板菜单 for "pane menu", and `history.unavailable.noSession` in ko says 이 패널 for "this pane". The other two 面板 hits are `crew.footer.aria`, where the English is "Open the crew overview" with no pane in it, so that one reads correctly as it stands.

## Toggle placement

Left where it is, per your reasoning about `DisplayPrefsContent` having exactly one caller.

## One addition you did not ask for

The branch had no `## [Unreleased]` line, which I had missed. `scripts/git-hooks/pre-commit` counts everything under `web/src/` as functional, so the branch owed one and would have been refused by your own guard. Added under `### Added`, with the lead-sentence shape `scripts/release-notes.ts` builds the release page from. Reword it however you like, or drop it if you would rather write that line yourself at merge.

Gate is green on the rebased tree: 241 files, 7490 tests, `tsc --noEmit` on both projects, and `oxlint --max-warnings 0`.

GitHub still lists the CI run as `action_required`, as it has on every push to this PR since it is from a fork, so it needs your approval to run.

AltanS:
Thanks @waynehoover, the branch is on main and the PR is closed by the merge. Your five commits are kept as they are, the changelog line credits you, and ADR 0002 now carries an addendum recording the override you built, with ADR 0047 pointing at it as the seam for an agent whose answer follows the terminal's background. I checked the switch on a phone-width build of main before pushing: the row sits inside the sheet without clipping, the choice survives a reload, and picking the agent's own answer removes the entry. This ships in the next release.

PR #244, opened 2026-09-19
feat(harness): Tier-1 + Tier-2 Muse adapter with native dialog UI

jpcarranza94:
Drives Muse panes end to end: idle/working/done chrome, the guarded reply path (type-then-verify against the draft box, paste-token supplement for per-line collapse), and four dialogs as native phone UI: shell approval, single-select, multi-select with review screen, workspace trust.

Evidence: 18 sanitized PTY captures (Muse 1.3.0) + live key probes. Digits move the pointer in Muse multi-select, so the neutral multi-select model gains a digit | pointer choreography; Claude fills digit, no behavior change. Signatures run question to dialog end with no transcript lookback (live spinner timers would churn it). ADR 0048 records the six decisions; MUSE_DIALOG_NOTES.md the choreography recipes.

Verified: new + touched suites 604 passed; full web suite 6799 passed (one flaky notify-prefs ordering failure, passes alone); lint 0 errors; typecheck and build clean.

Merge gate: maintainer live-verification of the send path against a real pane.


AltanS:
Thanks @jpcarranza94, this is careful work, and it lands on main for 1.11.0. I kept both of your commits with your authorship, but I cherry-picked them instead of merging the branch, so GitHub will not show this PR as merged and I will close it by hand. On top I added one commit. A lifted dialog now keeps the rows it asks about, so the approval shows the `$` command and argv and the trust prompt shows the workspace path, because a tap must show what it agrees to. A question or checkbox dialog lifts only while the composer under it is empty, and the review screen only while the `Request user input` header sits right above it, because text Muse prints in a reply looks the same and must never get buttons. The review's cancel button says `Interrupt turn`, since that is what Muse does with it. The raw terminal view lifts nothing, the pane strip uses the Muse mirror colors, and your ADR became a follow-up section in 0047 with your notes moved to `muse/DIALOG_NOTES.md`. You get credit in the changelog. I have no Muse install, so please run main against Muse when you can and open an issue for anything that looks wrong: approve a command with Allow once, Always allow (then run it again and see no dialog) and Abort, each once with the pointer moved first; pick option 3 of a question with the pointer on 1; open a note with Tab and see the buttons go; toggle a checkbox, including the row under the pointer, then Submit and confirm on the review screen, and once use Interrupt turn; trust a new folder; send a short reply, a long wrapped one and one over 1200 characters; ask Muse to print a fake approval, question and review screen in its reply and see no buttons; and check the raw terminal and the light theme. A short screen recording of the review Submit and Interrupt turn would help most.

Issue #249, opened 2026-09-20
Codex native composer leaves a background strip above the status bar


Femoon:
### What happened, and what did you expect?

When viewing a Codex pane in Collie on an iPhone, a partial dark background strip remains above the status bar. It appears to be part of the native Codex input area.

Steps to reproduce:
1. Open a Codex session in Herdr.
2. View the pane in Collie on an iPhone.
3. Wait until Codex is idle and scroll to the bottom.

**Expected:** Only Collie's own composer should be visible, without remnants of the native input area.

**Actual:** A dark background strip remains above the status bar. This does not appear in Claude Code panes.

Environment: Collie 1.10.2 (also reproduced on 1.10.1), Herdr, Codex, iPhone.

<img src="https://raw.githubusercontent.com/Femoon/collie/981184d0694e1ffbc68ea41701fec837c3bf882b/codex-before.jpg" width="320" alt="Codex pane with a persistent background strip highlighted in red above the status bar" />

### Output of `collie doctor --json`

Collected from the current local patched installation, not the original 1.10.1 installation shown in the screenshot.

<details>
<summary>Full diagnostic output</summary>

```json
[
  {
    "check": "collie",
    "status": "ok",
    "detail": "v1.10.1+166cabdc-dirty · macos-arm64",
    "remedy": null
  },
  {
    "check": "config-file",
    "status": "skipped",
    "detail": "home /Users/femoon/.collie/config.toml (absent) · instance /Users/femoon/.config/collie/config.toml (absent) — no config file, the env decides everything",
    "remedy": "`collie config init`"
  },
  {
    "check": "web-dist",
    "status": "ok",
    "detail": "25 entries, index.html present",
    "remedy": null
  },
  {
    "check": "path-link",
    "status": "ok",
    "detail": "not linked — `collie link` would publish /Users/femoon/.local/bin/collie → /Users/femoon/.local/share/collie/current/bin/collie",
    "remedy": null
  },
  {
    "check": "install",
    "status": "ok",
    "detail": "binary install, version 1.10.1-codex-spacing-20260920 at /Users/femoon/.local/share/collie (5 previous kept)",
    "remedy": null
  },
  {
    "check": "versions",
    "status": "ok",
    "detail": "current 1.10.1-codex-spacing-20260920 · 5 retained (1.10.1, 1.10.1-codex-padding-20260920, 1.10.0, 1.9.1, 1.6.0)",
    "remedy": null
  },
  {
    "check": "update-source",
[...]

Issue #245, opened 2026-09-19
Codex Astra TUI rendering artifacts in Collie


dondiegorivera:
### What happened, and what did you expect?

When viewing a Herdr session through Collie, Codex Astra’s sparkling/animated lower section causes severe rendering artifacts (see screenshot).

Steps to reproduce

1. Run Codex with an Astra model in Herdr.
2. Open the Herdr session in Collie.
3. Observe the sparkling/animated area of the Codex TUI.

<img width="630" height="1368" alt="Image" src="https://github.com/user-attachments/assets/12f70db4-ef43-4bcf-bcca-bf09a8aa927f" />

### Output of `collie doctor --json`

```json
Its a visual bug
```

### If collie will not start at all

```shell

```

### If this is an install problem

```shell

```

AltanS:
Thanks @dondiegorivera, this reproduces clearly from the screenshot. The black bars each start at a column and run to the right edge, which is a torn frame: Codex repaints the starfield continuously and the mirror caught a row half written. There is a second risk in the same image. Collie finds Codex's input box by locating the status row at the bottom and walking up at most two blank rows to the `› ` prompt, and here about six starfield rows sit between them. Depending on what those rows contain, that either refuses sends from the phone or reads the sparkles as part of your draft. Could you run `scripts/capture-fixture.sh <paneId> astra-composer` against that pane, with pane ids from `GET /api/snapshot`, and attach the file? The repository is public, so read it through for anything private first. With the real bytes I can fix both halves and pin them against a fixture.

AltanS:
Thanks @dondiegorivera, the next release fixes the part of this that stopped Collie from working with Astra. Our own capture of a gpt-6-astra pane showed the sparkles as single braille glyphs, each in its own grey. Collie now reads past them: sends from the phone go through with sparkle rows around the input line, and an empty input line no longer shows up as a stranded draft. The torn rows in your screenshot are a different matter. Codex repaints the starfield all the time, and the mirror sometimes catches a row halfway through a repaint, which Collie cannot prevent without running a terminal emulator. If Codex has a setting that turns the starfield off, that removes the tearing too. You get credit in the changelog, and I am closing this now.

PR #250, opened 2026-09-20
fix(codex): remove native composer background padding

Femoon:
Affected versions: Collie 1.10.1 and 1.10.2 (reproduced on both).

Codex paints a blank background row above its native `›` composer. Collie removes the prompt but leaves that padding visible as a dark strip above the status bar. This change includes adjacent blank rows with the prompt's background fill in the composer band, without changing prompt detection or draft binding.

Refs #249. Related context: #245 addresses Astra starfield handling; this change covers solid background padding.

Also align the transcript-to-status-bar spacing with Claude Code by trimming trailing separator blanks. Interior paragraph spacing is preserved.

### Screenshots

The after screenshot is from a local 1.10.1 backport of this change.

| Before | After |
| --- | --- |
| Before: native Codex composer background remains above the status bar | After: native composer background removed and bottom spacing aligned |

### Validation

- Root and web typechecks and lint pass.
- Full frontend suite: 7444 passed, 34 todo.
- Codex and Claude chrome tests: 693 passed, 5 todo.
- Playwright: 6 tests passed across Chromium phone/tablet and iPhone WebKit, in light and dark themes. Includes a direct Codex/Claude spacing comparison (within 1 CSS pixel).
- The local stable backport is deployed on three Collie nodes. Health, crew connectivity, served assets, and unchanged peer pane identities were verified.
- Full backend suite (`bun test ./bridge ./cli ./scripts`): 5034 passed, 0 failed across 164 files on the current PR head. Run on macOS arm64 with isolated Bun 1.4.2 and a non-symlink `TMPDIR`; no tests excluded and no source changes. This avoids the old Bun 1.3.14 hang and the macOS `/var` versus `/private/var` test-path mismatch.

**Base branch**

- [x] Opened against `main`.

**Checks**

- [x] `bun run typecheck` (root **and** `web/`), `bun run lint`, `bun test ./bridge ./cli ./scripts`, `cd web && bun run test` all pass locally with the test environment described above.
- [x] CHANGELOG/version exemption applies: this is a fork PR; release files are unchanged.


AltanS:
Thanks @Femoon, this goes in as it stands. The one line in `stripChrome` matches what Claude's pipeline already does, so a single change covers both cases. The unit tests read a real ANSI capture rather than a text fixture, and the negative control is the case I would otherwise have had to ask for. The browser spec goes in too, because it fails on the current code and passes on yours. Your authorship stays, the changelog line credits you and names #249, and the PR closes when the branch merges.

PR #253, opened 2026-09-20
feat(web): serve the app under a base path

edwinhu:
Collie assumes it is mounted at the origin root: `scope`/`start_url`/`id` are `/`, the router
basename is `/`, and `apiFetch`, the font URLs and the push helpers all build root-absolute paths.
Put it behind a proxy that gives it a path and every one of those leaves the mount.

Concretely, with `tailscale serve` mapping `https://host/collie` to the bridge on `127.0.0.1:8790`
(one tailnet node serving several apps by path prefix — `tailscale serve` forwards the prefix and
has no strip option):

```
https://host/collie/api/health   200  text/html   <- the SPA fallback, not the API
https://host/api/health          404
http://127.0.0.1:8790/api/health 200  {"ok":true,...}
```

This adds `COLLIE_BASE_PATH`. vite reads it as `base`; the router, the manifest (`id`,
`start_url`, `scope` and the icon `src`s), the font URLs, the push helpers and the service
worker then resolve against `import.meta.env.BASE_URL` instead of the origin, so index.html,
the assets, the API and a notification tap all stay inside the mounted prefix.

**Unset — the default — `BASE_PATH` is `/`, the prefix is empty, and every resolution is
byte-for-byte the one on main today.** Nothing changes for a root-mounted deployment.

Checks: root `typecheck`, `web typecheck`, `lint`, and `check-payload-links.test.sh` pass.
`cd web && bun run test` matches the baseline on stock `main` exactly in my checkout (same
159 fail / 7296 pass, a local jsdom issue unrelated to this branch).

No version bump or CHANGELOG edit, per CONTRIBUTING's fork rule. Suggested CHANGELOG entry:

- **Collie can be served under a path instead of the origin root.** A deployment behind a proxy that mounts it at `https://host/collie/` sets `COLLIE_BASE_PATH`, and the router, manifest, fonts, push helpers and service worker resolve against the mount rather than the origin. Unset, every resolution is unchanged.


AltanS:
Thanks @edwinhu, the branch is merged as it stands and the mount now works from the release build too. Your map of every place the app assumed the origin root was the whole job, and it is what the follow-up commit is built on. The one thing that changed is where the value lives: the release payload ships web/dist prebuilt, so a build-time base could never reach a binary install, and a checkout lost it at the next collie update. COLLIE_BASE_PATH is now read by the bridge at runtime, the bundle's internal references are relative and the bridge resolves the shell to the mount when it serves index.html, so one build serves any path. One correction to the description, because it matters for the docs: tailscale serve strips the mount before it proxies, the source does a StripPrefix on every path except a bare slash, and the bridge now also accepts a request that still carries it, so either proxy style works. collie serve publishes the door at the mount and doctor names it. You are credited in the changelog and in ADR 0052, and it ships with the next release.

